PR #19, opened 2021-01-06
VT DD list add routes

veeteeran:
## Description

This PR adds three components: one for list view, add item view, and navigation.

### Add dependencies for react router dom
`npm install --save react-router-dom`

### Navigation tabs
Clicking navigation links updates to the list or add item views. The text in tab displays in bold and the background color changes.

### Route NavLink Switch
Components of react router dom. Route renders parts of the UI, NavLink allows navigation without a full refresh and also styling attributes, Switch renders a route exclusively.
More documentation on `react-router-dom` [here](https://reactrouter.com/core/guides/quick-start)

## Related Issue

Closes [#3](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/3)

## Acceptance Criteria

- [ ] The following has been added as a project dependency: `react-router-dom`
- [ ] Links are present and persistent at the bottom of the app: one for the “list” view, the other for the “add an item” view
- [ ] When one of the links is clicked, the browser URL updates to represent the current view and a matching view component is displayed
- [ ] Whichever view is selected, the corresponding link should display in bold text

## Type of Changes

|  ✓ | :link: Update dependencies |

## Updates

list_view
add_view

* From your terminal, pull down this branch with `git pull origin VT-DD-list-add-routes` and check that branch out with `git checkout VT-DD-list-add-routes`
* Then `npm install --save react-router-dom`  to install the newly added dependencies locally and `npm start` to launch the app.
* Or as Lauren suggested but dark mode!
Screen Shot 2021-01-05 at 7 37 17 PM

* Click the list or add an item button change views


LINDGRENBA (approved):
Looks great! It looks like all of the AC have been met and the detail of your PR is excellent, which made it very easy to test and review. I like that you added the effect of a background on whichever link was active. One very small thing that I noticed is that the `AddItem` component is structured with a div while the `List` component is just the `<h1>`
```
<div>
      <h1>Add an Item</h1>
 </div>
```
`return <h1>List</h1>;`

I don't think this impacts functionality, I just noticed the difference. Great job you two!!!

src/components/Navigation.js
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <nav>
+      <Link
+        to="/list"
+        className="list-link"
+        onClick={toggleBold}
+        style={{ fontWeight: listFontWeight }}
+      >
+        List
+      </Link>
```
laurenmbeatty, line 27: I'm going to make a suggestion because I think it will make your lives easier. : )

`NavLink` from react-router has a built-in `activeClassName` for when the `to` prop matches the url. That way, if you are on the `/list` route and you refresh the page, that link will still be the `active` link, and you can give it the bold font and background color based on the `activeClassName`.

https://reactrouter.com/web/api/NavLink

laurenmbeatty:
Did you all branch off of Brittany and Rachel's branch? Then you can submit the pull request to merge into their branch:

1. Click edit at the top:
image

2. change the base branch
image

src/App.css
```diff
@@ -20,3 +25,25 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-bottom: 2%;
+}
+
+nav a {
+  text-decoration: none;
+  color: black;
+  font-size: 1.1rem;
```
MikeBLambert, line 40: Nice use of rem!

src/App.css
```diff
@@ -20,3 +25,25 @@
...
+  padding: 3% 0;
+  margin: 2%;
+  width: 30%;
```
MikeBLambert, line 43: Percents can be super tricky for things like width. If you take a look at the app in mobile view (see the circled sections in the screen shot for how to view as mobile) , notice how the button heights end up different b/c the content of one button has to wrap, while the other one is just fine in smaller screens.

Now, the tough thing here that you may want to get input from the team on is really the question of "How do we want these buttons to look on the smaller screens?" I love the large font personally and would hate to make the text smaller to solve the issue. Should we make the buttons have a fixed width that is large enough to accommodate both texts? Do we want both buttons to be taller? These are questions that I usually reach out to my team's designer about, but a lot of teams don't have designers and in that case, it's perfectly fine and usually best to reach out to other members of the team when making design decisions like this.

image

src/components/Navigation.js
veeteeran: I've made changes locally and want to review with Dee before pushing to the remote. Thanks for the tip! It simplifies things a lot.

src/App.js
```diff
@@ -1,11 +1,26 @@
...
+import AddItem from './components/AddItem';
+import List from './components/List';
 
 function App() {
   return (
     <div className="App">
+      <Navigation />
       <Test />
+
+      <Switch>
```
MikeBLambert, line 15: Somebody read the documentation! Nice job.

src/App.css
```diff
@@ -1,5 +1,10 @@
 .App {
   text-align: center;
+  display: flex;
+  flex-direction: column-reverse;
```
MikeBLambert, line 4: Flex! Great choice. I love flex! 

I see that you used `flex-direction: column-reverse`. What kinds of implications would using `column-reverse` vs `column` having moving forward?

rachelrohrbach (approved):
This looks great! AC has all been met and I liked that you linked to some react documentation in the PR as well.

src/components/Navigation.js
DeeDowns: I just learned something new, thanks

laurenmbeatty: I learned about `NavLink` from a former Collabie!

src/components/Navigation.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+
+const Navigation = () => {
+  return (
+    <nav>
+      <NavLink to="/list" activeClassName="list-link">
+        <div className="list-link-text">List</div>
+      </NavLink>
+      <NavLink to="/add-item" activeClassName="add-link">
```
laurenmbeatty, line 10: Great! Now when you refresh, the link stays bold. I think the `activeClassName` could be the same for both links...keeps it simple. It would be a good idea to check the font color and the background color with this tool, to make sure it's accessible: https://webaim.org/resources/contrastchecker/

src/components/Navigation.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+
+const Navigation = () => {
+  return (
+    <nav>
+      <NavLink to="/list" activeClassName="active-link">
```
chrisdopuch, line 7: Nice use of activeClassName 👍 

Minor suggestion: you could save this string as a `const` and reuse it in both places. That will be even more useful if you add several more link items!

src/App.css
veeteeran: Hi @MikeBLambert We've fixed the odd box sizing issue and for now when the screen width is less than 900 the boxes resize to 50px x 50px. The text still wraps though. I'll consult the team as to see how we want to handle the nav buttons on a mobile device.

Screen Shot 2021-01-06 at 7 21 07 PM

src/App.css
```diff
@@ -1,22 +1,42 @@
...
+}
+
+@media screen and (max-width: 900px) {
+  nav a {
+    width: 50px;
+    height: 50px;
+  }
+
+  .list-link-text {
+    margin-top: 12px;
```
chrisdopuch, line 35: Manually specifying a pixel value for margin is a good simple start for centering the text vertically 👍 

But, what if the height of our container changed? Then we would also need to update this pixel value to match.
Fortunately, flexbox gives us a handy way to center content vertically within a container - wiithout needing to specify the pixel height manually!

See this example where I removed the margin, and replaced it with flexbox layout:


Feel free to keep this as it is if you want to, it does work fine for now. But, if you want to "flex" your flexbox muscles, this would be a good thing to attempt and learn! 😃

chrisdopuch (approved):
Overall looking good to me 👍 
Worked great on my machine locally
Nice job y'all! 🚀

src/App.css
MikeBLambert: And the pun of the PR award goes too... 😆

MikeBLambert (approved):

laurenmbeatty (approved):
Gorgeous! Nice work @DeeDowns and @veeteeran!

laurenmbeatty:
Oh, and I loved that you all branched off of Rachel and Brittany's work...very smart, and now you won't have to manage merge conflicts.

PR #20, opened 2021-01-08
Bl rr create firebase hook

LINDGRENBA:
## Description
This code removes the `useEffect` hook from the `Test.js` component and abstracts it out into a custom `useFirestore` hook. This update makes the code more reusable as `useFirestore` can now be called from any component. `useFirestore` takes a string as an argument which is the value of the collection to be accessed. For example `useFirestore('tests');` will access the `'tests'` collection.

## Related Issue
[#2](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/2)

## Acceptance Criteria
Code should still meet all AC from issue [#2](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/2)
1. The following have been added as project dependencies: firebase & react-firebase-hooks
2. Make a change in the Firestore database and it shows up in the app
3. Make a change in the app and it shows up in the Firestore database

## Type of Changes

|   ✓ | :hammer: Refactoring       |

### Before
does not apply

### After
does not apply 

* From your terminal, pull down this branch with `git pull origin BL-RR-create-firebase-hook` and checkout the branch with `git checkout BL-RR-create-firebase-hook`.
* Run `npm install` to install any new dependencies, then run `npm start` to launch the application.
* Click on the `Click Me!` button and make sure that a new `Hello World!` message appears.

### DIFF
--- a/src/hooks/useFirestore.js
+++ b/src/hooks/useFirestore.js
@@ -5,22 +5,18 @@ const useFirestore = (collection) => {
   const [docs, setDocs] = useState([]);
 
   useEffect(() => {
-    const unsubscribe = db.collection(collection).onSnapshot(
-      (snapshot) => {
-        let documents = [];
-        snapshot.forEach((doc) => {
-          documents.push({ ...doc.data(), id: doc.id });
-        });
-        setDocs(documents);
-        return () => {
-          unsubscribe();
-        };
-      },
-      [collection],
-    );
-
-    return { docs };
-  });
+    const unsubscribe = db.collection(collection).onSnapshot((snapshot) => {
+      let documents = [];
+      snapshot.forEach((doc) => {
+        documents.push({ ...doc.data(), id: doc.id });
+      });
+      setDocs(documents);
+      return () => {
+        unsubscribe();
+      };
+    });
+  }, [collection]);
+  return { docs };
 };
 
 export default useFirestore;
--- a/src/hooks/useFirestore.js
+++ b/src/hooks/useFirestore.js
@@ -1,0 +1,26 @@
+import { useState, useEffect } from 'react';
+import { db } from '../lib/firebase';
+
+const useFirestore = (collection) => {
+  const [docs, setDocs] = useState([]);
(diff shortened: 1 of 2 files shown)


src/hooks/useFirestore.js
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from 'react';
+import { db } from '../lib/firebase';
+
+const useFirestore = (collection) => {
```
laurenmbeatty, line 4: 🤩  Wow! This is very sophisticated.

laurenmbeatty (approved):
Great refactor. Love it.

MikeBLambert (approved):
This is AMAZING! Your task was just to connect to the database via some code that was effectively just going to be thrown away down the road. Without going out of scope of the ticket, you were able to not only connect to the database, but also do so by creating a utility that may end up being super duper useful for the team as you move forward.  GREAT JOB!

src/hooks/useFirestore.js
```diff
@@ -0,0 +1,22 @@
...
+  const [docs, setDocs] = useState([]);
+
+  useEffect(() => {
+    const unsubscribe = db.collection(collection).onSnapshot((snapshot) => {
+      let documents = [];
```
chrisdopuch, line 9: Fun fact - you could actually still use a `const` here, even though you're pushing new items onto the array. Const just makes the _reference_ or variable name immutable, but not the data that the reference/name points to.

So like `documents = someOtherList` would throw an error with `const` but `documents.push({}` would not.

chrisdopuch (approved):
Very nice work 👍 
This hook will make things easier down the road for reusing the firebase connection code in other components 🎉

src/hooks/useFirestore.js
LINDGRENBA: Ah, good to know! I'll make that change before merging.

LINDGRENBA:
Thank you everyone for all of the great feedback!

PR #21, opened 2021-01-11
Add comment with issue AC to project.

LINDGRENBA:
## Description
This PR adds the `PropTypes` dependency and allows the user to click on a button and will utilize the script in `src/lib/tokens.js` to generate a token and save it to `Local Storage`

## Related Issue
[#4](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/4)

## Acceptance Criteria
- [x] For users who do not already have a token/list, a button or link exists on the home screen that allows them to create a new list
- [x] Clicking the button/link generates a new token and saves it to `Local Storage`
- [x] Once the token has been created and saved, the user is redirected to the "list" view
- [x] For users who do have a token already saved in `Local Storage`, they should be shown/redirected to the "list" view

## Type of Changes

## Updates
From the home screen we created a `Create a New List` button to generate a token and save the token to `Local Storage`. Once created the user is redirected to the list view. If the token already exists in `Local Storage` the user is redirected to the list view.

### Before

Screen Shot 2021-01-06 at 11 49 41 AM

### After

Screen Shot 2021-01-12 at 6 10 00 PM

* Otherwise from your terminal, pull down this branch with `git pull origin BL-VT-generate-token-new-list` and check that branch out with `git checkout BL-VT-generate-token-new-list`
* install new dependencies with `npm install`
* Click the `Create a New List` button to generate a token and go to List view
* To see the token in `Local Storage`, open developer tools and click on `Application` tab. Find the token under the `Local Storage` dropdown menu on left side of the screen

Screen Shot 2021-01-12 at 6 04 41 PM


src/App.js
```diff
@@ -1,13 +1,21 @@
...
 
 function App() {
+  const [token, setToken] = useState(localStorage.getItem('userToken') || '');
+
+  const handleClick = () => {
+    const token = getToken();
+    localStorage.setItem('userToken', token);
+    setToken(token);
+  };
+
```
LINDGRENBA, line 18: @veeteeran  and I discussed creating a `useLocalStorage` custom hook here. Do others agree this would be a good update to make the code more reusable or would this be adding unnecessary complexity? If it's a good idea, should the custom hook set and get the token, or just set it.

src/App.js
```diff
@@ -1,13 +1,21 @@
...
 import './App.css';
-import Test from './components/Test';
 import Navigation from './components/Navigation';
 import AddItem from './components/AddItem';
 import List from './components/List';
 import Welcome from './components/Welcome';
+import getToken from './lib/tokens';
 
 function App() {
+  const [token, setToken] = useState(localStorage.getItem('userToken') || '');
```
LINDGRENBA, line 11: the conditional statement sets the `token` default value to our token from `localStorage` if it exists or to an empty string if no token has been generated yet.

src/App.js
```diff
@@ -20,11 +28,10 @@ function App() {
         </Route>
 
         <Route path="/">
-          <Welcome />
+          {token ? <Redirect to="/list" /> : <Welcome onClick={handleClick} />}
```
LINDGRENBA, line 31: This will redirect to `list` if the initial default value of `token` is a token from `localStorage`

src/App.js
rachelrohrbach: I'm not sure if we need a custom hook. If we can ensure that the user token exists before the user goes to the Add Item page then we can pull the user token directly from localStorage.

src/App.js
rachelrohrbach: The rendering for the Welcome component is inside the "/" route, so if a user navigates directly to Add Items or List there is potential that the user token is not already set in localStorage.  So I would suggest that we might want to render Welcome instead of going through the routes logic when no user token exists.

rachelrohrbach (approved):
Great job! :clap: It looks like you met the AC, and the detailed instructions for testing were great! I had forgotten how to look at localStorage, so I really appreciated the specific steps with the screenshots. I also found the documentation and videos that you guys found when researching localStorage to be helpful recaps and some new lessons :smile:

src/App.js
rachelrohrbach: The rendering for the Welcome component is inside the "/" route, so if a user navigates directly to Add Items or List there is potential that the user token is not already set in localStorage. So I would suggest that we might want to render Welcome instead of going through the routes logic when no user token exists.

src/App.js
LINDGRENBA: Okay, awesome!

src/App.js
LINDGRENBA: If we just render `<Welcome />`, then would we do something like `{ token && <Redirect to="/list" />"` underneath the `<Welcome />` component or could the `<Redirect /> be anywhere within the `App` component? Or is the suggestion that the logic for managing the `token` state and generating the token and saving it to `localStorage` should just be inside of the `<Welcome />` component? I just want to make sure I'm on the same page! 👍

rachelrohrbach: I'm thinking that we should not render the components that require a user token until one exists.  The logic that renders the components that depend on user token are handled by the Switch component, so do not render Switch component until the user has a token.

src/App.js
```diff
@@ -1,13 +1,21 @@
...
+
+  const handleClick = () => {
+    const token = getToken();
+    localStorage.setItem('userToken', token);
+    setToken(token);
+  };
+
   return (
     <div className="App">
       <Switch>
```
rachelrohrbach, line 21:
```suggestion
      <Switch>
        <Route path="/list">
          {token ? <List /> : <Welcome onClick={handleClick} />}
        </Route>

        <Route path="/add-item">
          {token ? <AddItem /> : <Welcome onClick={handleClick} />}
        </Route>

        <Route path="/">
          {token ? <Redirect to="/list" /> : <Welcome onClick={handleClick} />}
        </Route>
      </Switch>
```

rachelrohrbach, line 21:
```suggestion
      <Switch>
        <Route path="/list">
          {token ? <List /> : <Welcome onClick={handleClick} />}
        </Route>

        <Route path="/add-item">
          {token ? <AddItem /> : <Welcome onClick={handleClick} />}
        </Route>

        <Route path="/">
          {token ? <Redirect to="/list" /> : <Welcome onClick={handleClick} />}
        </Route>
      </Switch>
```

rachelrohrbach: I think we should render the Welcome component if the user does not already have a user token. I think there are probably ways we could clean this up so we don't duplicate code.

laurenmbeatty: @rachelrohrbach I like your suggestion! Instead of the welcome component in those top two ternaries (lines 23 and 27), I think a `<Redirect to="/" />` would do the trick. I imagine it could be even fancier, but I _think_ that would get the job done, so to speak. Otherwise, without a token, the user could be at the `/list` in the url, but it would have the `Welcome` screen.

src/App.js
laurenmbeatty: Yep, another way would be to pass the token down as a prop to the components that need it.

MikeBLambert: I always find it super tough to decide when to add an "abstraction", which is what a custom hook would be here. I think that most devs (including me) jump to making abstractions because, well, it's fun! It's also super tempting to want to pre-optimize our code.

In response to the traditional idea of "DRY Programming" DRY (Don't Repeat Yourself), there is a new trend among the cool kids of programming these days called "AHA Programming" (Avoid Hasty Abstraction). Here's a really great blog post about it from one of the React superstars - https://kentcdodds.com/blog/aha-programming 

Long story short, lately I've been trying to wait until I've repeated some code a bunch of times before I create an abstraction around it. By waiting, it usually gives you a much better idea of what the really problem that the abstraction is solving and how best to write it.

src/App.js
LINDGRENBA: Awesome! Thank you Rachel. This is a great suggestion. I'll get together with Viet on this! Great suggestion and we can check in with our mentors regarding ways to reduce duplicate code!

src/components/Welcome.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 
-const Welcome = () => {
+const Welcome = (props) => {
```
MikeBLambert, line 3: Nit-pick: Adding propTypes are a great way to warn if forgetting to add pass a prop to a component or if passing the wrong data type as a prop. They're also a great way to provide documentation to your code. https://reactjs.org/docs/typechecking-with-proptypes.html

src/App.js
MikeBLambert, line 11: impressive use of the logical || operator!

MikeBLambert (approved):
I love the PR conversations going on. Really great work and great collaboration!

src/App.js
LINDGRENBA: Thanks for the feedback! Epicodus had us use PropTypes, but I haven't seen them used in some of the recent tutorials I've watched, so it's good to get a reminder on why they're helpful! And thank you for linking to some additional resources for further learning 👍

MikeBLambert: In my opinion, the #1 reason to use prop-types is for documentation purposes. It's always really useful when I go into a component and see something that looks like:

```
Component.propTypes = {
   // Super useful comment explaining the exact purpose of the prop
   onClick: PropTypes.func.isRequired
}
```
Whenever I see well documented prop-types, I usually know that it was written by someone who was really trying to be helpful to other devs :-)

src/App.js
veeteeran: Thanks @MikeBLambert that was a great take on abstraction. I haven't heard about AHA before and I want to be a cool kid! 😆 I feel like everyone tcl-17 is planning ahead. If we were to build out a task list with columns `urgent`, `do soon`, `great if we have time` this custom hook could be put in the last column for now. Didn't @laurenmbeatty have a slide like this during our first sync? 🤔

src/App.js
veeteeran: Thanks @rachelrohrbach and @laurenmbeatty for the feedback! Thinking about how we implement this 🤔

src/components/Welcome.js
```diff
@@ -1,11 +1,18 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-const Welcome = () => {
+const Welcome = (props) => {
+  const { onClick } = props;
```
laurenmbeatty, line 5: Nice destructuring, I like it. Another way to do this would be to destructure inline on line 4:

`const Welcome = ({onClick}) => {`

src/App.js
```diff
@@ -1,30 +1,39 @@
...
+import { Route, Switch, Redirect } from 'react-router-dom';
 import './App.css';
-import Test from './components/Test';
 import Navigation from './components/Navigation';
 import AddItem from './components/AddItem';
 import List from './components/List';
 import Welcome from './components/Welcome';
+import getToken from './lib/tokens';
+// import TestWelcome from './components/TestWelcome';
+// import CreateNewList from './components/CreateNewList';
```
laurenmbeatty, line 10: Can get rid of the commented out code.

src/App.js
```diff
@@ -1,30 +1,39 @@
...
         <Route path="/add-item">
-          <AddItem />
+          {token ? <AddItem /> : <Redirect to="/" />}
         </Route>
 
         <Route path="/">
-          <Welcome />
+          {token ? <Redirect to="/list" /> : <Welcome onClick={handleClick} />}
+          {/* <TestWelcome />
+          {token ? <Redirect to="/list" /> : <CreateNewList onClick={handleClick} />} */}
```
laurenmbeatty, line 33: Same here, can get rid of the commented out code.

laurenmbeatty (approved):
This looks great! I would suggest removing the commented out code, and even getting rid of the Test component completely? I don't think we need it anymore.

src/App.js
LINDGRENBA: It's definitely a challenge to find the right balance between DRY and AHA. I like the idea of waiting until you have a better idea of how best to solve the problem!

src/App.js
LINDGRENBA: Okay! We've implemented those changes. Yay team!

src/App.js
LINDGRENBA: I feel like I've learned so much working on this issue and reading through everyone's feedback. Thanks team!

src/components/Welcome.js
LINDGRENBA: Nice clean solution @laurenmbeatty !

src/App.js
LINDGRENBA: 👍

src/App.js
LINDGRENBA: 👍

src/App.js
```diff
@@ -1,30 +1,40 @@
...
 import './App.css';
-import Test from './components/Test';
 import Navigation from './components/Navigation';
 import AddItem from './components/AddItem';
 import List from './components/List';
 import Welcome from './components/Welcome';
+import getToken from './lib/tokens';
 
 function App() {
+  const [token, setToken] = useState(localStorage.getItem('userToken') || '');
```
DeeDowns, line 11: So this is what we will need so that the user can have a list?

src/components/Welcome.js
```diff
@@ -1,11 +1,17 @@
...
   return (
     <div>
       <h1>Welcome</h1>
+      <button onClick={onClick}>Create a New List</button>
     </div>
   );
 };
+/* Checks that onClick is a function and is required as a prop */
+Welcome.propTypes = {
+  onClick: PropTypes.func.isRequired,
```
DeeDowns, line 14: I've never used PropTypes before, is this like a type checker?

DeeDowns (approved):

src/components/Welcome.js
chrisdopuch: PropTypes is kind of like a type-checker, yeah. The only difference is that PropTypes checks for the types at _run time_ while things like TypeScript or Flow check them at _build time_. If you use the wrong type of prop, PropTypes will log a warning in your console while the app is running, but it will still "work" (as long as the prop type mismatch doesn't break your code!). TypeScript or Flow would actually prevent you from deploying code which contained such a bug in the first place. 
It's a good incremental step in this direction. PropTypes is a great way to add this kind of thing without worrying about additional tooling or type syntax. 
More docs and info: https://reactjs.org/docs/typechecking-with-proptypes.html

chrisdopuch (approved):
Looks good y'all, worked nicely on my machine 👍

PR #22, opened 2021-01-12
Add comments to describe list item object

rachelrohrbach:
## Description

This PR will add an input form for the user to add new items to the user's shopping list including the name of the item, when the user will need to purchase it again and last purchased date (null if n/a). When the user clicks the submit button it will save the item to the db associated with the users token.

## Related Issue

#5 

## Acceptance Criteria

- [x] User is presented with a form that allows them to enter the name of the item and select how soon they anticipate needing to buy it again (Soon, Kind of Soon, Not Soon)
- [x] When the user submits the form, the item is saved to the database, associated with the user’s token
- [x] Along with the item name, an integer corresponding to the estimated number of days until next purchase is saved: 7 for “Soon”, 14 for “Kind of Soon”, and 30 for “Not Soon”
- [x] The last purchased date should be set to null initially (i.e. you can create an item without purchasing it)
- [x] Item names should be displayed as a simple, unordered list on the list view after they’re added

## Type of Changes

## Updates
* Created add item form with validation for the `itemName` input
* Sent form data to the `firestore` database, associated with the user's token (for now a placeholder, but eventually will be passed as props)
* From `List` component, use `useFirestore` hook to grab data, then mapped through to display item names as unordered list items
* Note: we are getting a warning whenever an item is added to the list. Initially it was fixed when we added an if statement to the `useFirestore` `useEffect` that checks the length of the docs, but once we added in form validation the warning came back. 
Screen Shot 2021-01-14 at 3 48 42 PM

### Before
Add Item View 
[
Screen Shot 2021-01-14 at 4 56 21 PM
](url)
List View (empty)
Screen Shot 2021-01-14 at 4 56 42 PM

### After
Add Item View 
Screen Shot 2021-01-14 at 4 57 06 PM
List View (with newly added item)
Screen Shot 2021-01-14 at 4 57 22 PM

Can preview and test features with the Netlify deploy-preview feature
Otherwise from the terminal, pull down the branch `git pull origin RR-DD-add-new-item-form`, then `npm start` to launch the app, navigate to the Add Item page, fill out form and submit to be redirected to the List page with your new item(s)


rachelrohrbach:
@DeeDowns I put together some preliminary brainstorming notes for our issue this week. 

## AddItem Component 
 Do you think we should create a separate AddItemForm component and then import this inside the AddItem component?  - just thinking about the amount of code required for the form

Our form needs to include:
1. Text input - itemName
2.  Radio Button input - itemFrequency
    * Soon 
           1. Default value = 7
           2.  Default checked
     * Kind of Soon
           1. Default value = 14
     * Not Soon
           1. Default value = 30
3. Submit button
    * Error handling - cannot submit until all fields are filled out (really only the item name will cause an error, so message like “Please enter an item”)

* We make a post request to the Firestore db sending the new item information
* After it’s submitted redirect user to the List page

## List Component
We need to make a get request to the Firestore db with the users token & return the list document, map through it and display data as unordered list of the item names

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+import { db } from '../lib/firebase';
+
+const initialFormState = {
+  itemName: '',
+  timeFrame: 7,
+  lastPurchased: null,
+};
 
 const AddItem = () => {
+  const token = 'user token';
```
rachelrohrbach, line 12: This is just a placeholder for the user token for now. @veeteeran @LINDGRENBA if you can pass the userToken as props to AddItem then we can grab it and use it.

src/hooks/useFirestore.js
```diff
@@ -5,6 +5,9 @@ const useFirestore = (collection) => {
   const [docs, setDocs] = useState([]);
 
   useEffect(() => {
+    if (docs.length > 0) {
```
rachelrohrbach, line 8: this is a solution that I found on Stack Overflow but it doesn't seem to be solving our unmounting warning. I found another solution that I think might actually solve it
```suggestion
    let isMounted = true;

    const unsubscribe = db.collection(collection).onSnapshot((snapshot) => {
      let documents = [];
      snapshot.forEach((doc) => {
        documents.push({ ...doc.data(), id: doc.id });
      });
      if(isMounted) setDocs(documents);
      return () => {
        unsubscribe();
      };
    });

    return () => { isMounted = false };
```

src/components/AddItem.js
rachelrohrbach, line 12: Token is just a placeholder right now. I know we discussed having token being passed as a props to AddItem so we could use the user token that's in local storage

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+  };
+
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    if (!formData.itemName) {
+      setError(`${event.target.name} field cannot be empty`);
+    } else {
+      db.collection(token).add(formData);
+      history.push('/list');
+      setFormData(initialFormState);
```
chrisdopuch, line 33: Hey @DeeDowns @rachelrohrbach I have a hunch that this might be the source of your error:
```
Warning: Can't perform a React state update on an unmounted component.
```
Notice how it says we are performing a state update (e.g. setFormData) _after_ the component has already been unmounted. This matches with the code I'm seeing here. Because this component unmounts when we change route and go to the /list page, when we try to update the state right afterwards we might end up trying to update state on an unmounted component. 
Maybe you could try moving this line of code above the `history.push` call?

chrisdopuch (commented):
So far so great! Worked well on my machine, and the code's coming along nicely.
I'll pause until this is peer-reviewed and finalized, feel free to ping me for a re-review!

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+  const [error, setError] = useState('');
+  console.log(error);
+
+  const handleFormChange = (event) => {
+    event.preventDefault();
+    setFormData({
+      ...formData,
+      [event.target.name]: event.target.value,
+    });
+  };
```
laurenmbeatty, line 24: Just a quick comment...I think. you may need to `parseInt()` the `timeFrame` value before sending it. If I remember correctly it's supposed to be saved as an int, and right now it's getting saved as a string, and I think even if you set it's value on the radio to an int, it will still send as a string...need to convert it to an `int`.

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+  lastPurchased: null,
+};
 
 const AddItem = () => {
+  const token = 'user token';
+  const history = useHistory();
+  const [formData, setFormData] = useState(initialFormState);
+  const [error, setError] = useState('');
+  console.log(error);
+
```
LINDGRENBA, line 17: When I test out the form in the browser, I'm not able to change the radio button.  I think that would be controlled with state as well, so you could do something like
`  const [radioValue, setRadioValue] = useState('Soon');`

then add a helper function to update radioValue
```
  const handleRadioChange = event => {
    setRadioValue(event.target.value);
  }
```
and then...

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+      history.push('/list');
+      setFormData(initialFormState);
+      setError('');
+    }
+  };
+
   return (
     <div>
       <h1>Add Item</h1>
+      <form onSubmit={handleSubmit}>
```
MikeBLambert, line 41: Nice job with the HTML form! form can be super complicated and it took me a while to realize/understand the full power of the form. Great job!

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+
+        <fieldset className="fieldset">
+          <legend>Time Frame</legend>
+          <label htmlFor="timeFrame"> How soon will you buy this again?</label>
+          <br />
+          <input
+            type="radio"
+            id="soon"
+            name="timeFrame"
+            defaultChecked
```
LINDGRENBA, line 63: and then ... adding the attribute of checked like this to each component
`checked={radioValue === "soon"}`
`checked={radioValue === "kinda-soon"}`
`checked={radioValue === "not-soon"}`
will cause that radio button to be checked only if the `radioValue` state is set to the corresponding value

LINDGRENBA (approved):

LINDGRENBA:
I love the collective list that pops up after submitting the form! I got a good laugh out of all of the items on there 👍

src/components/AddItem.js
chrisdopuch: I can't verify that this actually fixes the warning, so I'm not totally sure. Still, it might be worth doing anyways, just in case. I'd put it down as the last line of this function, so that setting formData and Error are both done before navigation away from this page.

src/components/AddItem.js
MikeBLambert, line 63: oh what?! I didn't know about the `defaultChecked`. That is awesome!

src/components/AddItem.js
MikeBLambert: One thing that you could do is actually go ahead and make it a prop right now. Then, in the parent component you could pass in the mock token
or...better yet, you could play around with `defaultProps` (https://www.codecademy.com/courses/react-101/lessons/this-props/exercises/defaultprops) and make the token be a defaultProp of "user token". Then, when we're ready to work with the real token the code will already be expecting the token to be a prop, you can switch the defaultProps.token to an empty string, and just pass in the real token! Just an idea...

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+            value="7"
+            onChange={handleFormChange}
+          />
+          <label htmlFor="soon"> Soon</label>
+          <br />
+          <input
+            type="radio"
+            id="kinda-soon"
```
MikeBLambert, line 71: Yes! Great job using `id` and `htmlFor`! 🙌 🙌

src/components/AddItem.js
veeteeran: Agree with @MikeBLambert about forms. I thought they were kind of boring before doing a project on that went in depth with forms. It made me notice when forms are difficult to use vs when they are easy to use.

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+        <label>
+          Item Name
+          <input
+            type="text"
+            placeholder="add your item here"
+            name="itemName"
+            value={formData.itemName}
+            onChange={handleFormChange}
+          />
+          {error && <span>{error}</span>}
```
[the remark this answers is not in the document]
laurenmbeatty: Agree with @MikeBLambert here. I also think adding `type="submit"` to the button will make it so the user can't click the button until the required field has value (I could be wrong about that, though).

src/components/AddItem.js
veeteeran: I ran across defaultProps while reading about propTypes [here](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values)

src/components/AddItem.js
veeteeran: While reading about propTypes I ran across documentation for [defaultProps](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values)

src/components/AddItem.js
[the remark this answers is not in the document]
laurenmbeatty: I'm wrong about that 😂   Still a good idea to add type="submit"

src/components/AddItem.js
veeteeran: I also noticed the radio button issue @LINDGRENBA brought up. Her solution sounds like something to try. You've got this!

src/components/AddItem.js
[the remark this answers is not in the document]
veeteeran: Good point @MikeBLambert. Adding `required` brings up a popup and styling around the input box. The way it's implemented now might need a little styling to take place to make the error message stand out. Do we want a default or something custom made? The current way works and I'm all for how the team wants to move forward with this.

veeteeran:
I like what you two have done so far. As @laurenmbeatty mentioned in Slack we'll get this done together, unless you two figure it out before office hours. Nice work! Keep it going!

src/components/AddItem.js
```diff
@@ -1,11 +1,102 @@
...
+  timeFrame: 7,
+  lastPurchased: null,
+};
 
 const AddItem = () => {
+  const token = 'user token';
+  const history = useHistory();
+  const [formData, setFormData] = useState(initialFormState);
+  const [error, setError] = useState('');
+  console.log(error);
```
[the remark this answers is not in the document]
DeeDowns: No we were testing to see if the error was being set, but forgot to take it out.

src/components/AddItem.js
DeeDowns: Yea, there was an issue with changing the radio buttons, we had to click it twice for the UI to change, but the state is changing (we checked in Components tab of chrome's react dev tools)

src/components/AddItem.js
DeeDowns: > Just a quick comment...I think. you may need to `parseInt()` the `timeFrame` value before sending it. If I remember correctly it's supposed to be saved as an int, and right now it's getting saved as a string, and I think even if you set it's value on the radio to an int, it will still send as a string...need to convert it to an `int`.

Would it be sent in as a number if the value were set like this:
```
<input
            type="radio"
            id="soon"
            name="timeFrame"
            defaultChecked
            value={7}
            onChange={handleFormChange}
          />
```
instead of `value="7"`

DeeDowns: So I tried the above changes and the value for timeframe is now a number so I think that works, but if parseInt is better we can change it

laurenmbeatty: I'm not 100% sure @DeeDowns...would be good to test it out, then it will be a simpler change, for sure.

laurenmbeatty: Oh perfect...no need for `parseInt` then.

src/components/AddItem.js
DeeDowns: > I can't verify that this actually fixes the warning, so I'm not totally sure. Still, it might be worth doing anyways, just in case. I'd put it down as the last line of this function, so that setting formData and Error are both done before navigation away from this page.

I moved them both about history.push but still get the warning

src/components/AddItem.js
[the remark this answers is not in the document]
DeeDowns: > One alternative to adding an error text here would be take advantage of the awesome power and might that is HTML5 and pass a `required` prop to the input field. As we're in a form and using the `onSubmit`, HTML5 will show a nice little popup when any `input` fields that are `required` are empty.
> Not sure if this will fit with the design that the team wants, but just wanted to throw that out there as an option.

I didn't know about that, I do like the little pop up and its a lot less verbose than setting another state variable
And it seems the warning we've been getting has gone away since I removed setting an error state and replaced it with the required prop.

src/components/AddItem.js
veeteeran: Nice! That's a clean solution!

src/components/AddItem.js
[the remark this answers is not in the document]
DeeDowns: No I lied, the waring is back 😅

src/components/AddItem.js
veeteeran: I think we may have to check the `checked` prop to a value of true or false like they do [here](http://react.tips/radio-buttons-in-reactjs/)

src/components/AddItem.js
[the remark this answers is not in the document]
rachelrohrbach: I like the little pop up and  I learned something new as well 😄

src/App.js
```diff
@@ -1,30 +1,40 @@
...
+    setToken(token);
+  };
+
   return (
     <div className="App">
       <Switch>
         <Route path="/list">
-          <List />
+          {/* Go to list view if token exists otherwise redirected to home */}
+          {token ? <List token={token} /> : <Redirect to="/" />}
```
rachelrohrbach, line 24: we're able to successfully pass the user token as props to both the List and AddItem components

src/components/AddItem.js
rachelrohrbach: We did end up having to use parseInt to pass it as a number to the db.

src/components/AddItem.js
rachelrohrbach: We removed the `event.preventDefault()` from the `handleFormChange()` and now the radio buttons are working as they should (able to select them).
And we added `onChange` and `checked` as attributes of the radio button inputs.

src/hooks/useFirestore.js
```diff
@@ -11,11 +11,12 @@ const useFirestore = (collection) => {
         documents.push({ ...doc.data(), id: doc.id });
       });
       setDocs(documents);
-      return () => {
-        unsubscribe();
-      };
     });
+    return () => {
```
rachelrohrbach, line 15: Return was incorrectly placed inside of the function passed to onSapshot. Moved to be returned from useEffect to unsubscribe when unmounting component.

src/components/AddItem.js
```diff
@@ -1,9 +1,88 @@
...
       <h1>Add Item</h1>
+      <form onSubmit={handleSubmit}>
+        <h2>Add your item</h2>
+        <label>
+          Item Name
+          <input
+            type="text"
+            placeholder="add your item here"
+            name="itemName"
+            required={true}
```
laurenmbeatty, line 45:
```suggestion
            required
```

laurenmbeatty (approved):
Nice...this is a really challenging story the second week. 💪

veeteeran (approved):

chrisdopuch (approved):
I tested it out on my machine and it's working great! 
I like the new validation error message that pops up with an empty text input 👍 
Nice work you two!!

src/components/AddItem.js
MikeBLambert: Yeah, for more complex forms there are actually entire tools built around handling form state. https://formik.org/ is one of the more popular ones. By no means am I suggesting using it here, but just I just wanted to share it in case you find yourself struggling with a complex form in a future project.

MikeBLambert (approved):
Great job on a very difficult issue! This works great for me and I love the solutions that you've come up with. Keep it up!

PR #23, opened 2021-01-18
Add comments with issue AC for preventing duplicate list items

### LINKED ISSUE #7: 6. As a user, I want to be alerted when I’m entering an item that is the same as something already on my list so I can avoid duplicate items.
AC:

* Show an error message if the user tries to submit a new item that has the exact same name as an existing item
* Show an error message if the user tries to submit a new item that has the same name as an existing item, where capitalization has been normalized and punctuation has been removed
* The user’s original input should be what gets saved in the database

rachelrohrbach:
## Description

This code will alert the user if an item already exists in the user's list.  

## Related Issue

Closes #7 

## Acceptance Criteria

- [x] Show an error message if the user tries to submit a new item that has the exact same name as an existing item
- [x] Show an error message if the user tries to submit a new item that has the same name as an existing item, where capitalization has been normalized and punctuation has been removed
- [x] The user’s original input should be what gets saved in the database

## Type of Changes

## Updates
**'tomato' is already in the list**

Screen Shot 2021-01-19 at 5 21 57 PM

**tomato spelled with special characters is recognized as 'tomato'**

Screen Shot 2021-01-19 at 5 22 39 PM

* Otherwise from your terminal, pull down this branch with `git pull origin RR-VT-prevent-duplicate` and check that branch out with `git checkout RR-VT-prevent-duplicate`
* From add item page enter an item that already exists to see the error message
* List view reflects exact user input


MikeBLambert (commented):
I just tried to use this to create my shopping list for tomorrow AM, which includes:
- Champagne
- Champagne
- Champagne
(Yes, lots of champagne...)
I believe that I should have seen an error message, but I may be doing something wrong (wouldn't be the first or the last time...)
If worst comes to worst, I guess I'll just buy a bunch of champagne! 

In looking at the code, I think that you're totally on the right track querying the DB for the item first and then blocking the creation of a new item if it already exists. We may want to take a closer look at that query in `handleSubmit`

rachelrohrbach:
@MikeBLambert I think I understand the issue. We're cleaning up the user input to use to compare, but what's actually being added to the db isn't "cleaned up" so it's not catching it. From research it seems that firebase is case sensitive, so if we also send a lowerCaseItemName as part of our formData and then compare that to cleanInput in the query we'd actually be comparing "apples to apples".   🍎🍏 

```
 const snapshot = await queryCollection
      .where('lowerCaseItemName', '==', cleanInput)
      .get();
```

`db.collection(token).add({
      ...formData,
      lowerCaseItemName: cleanInput
    });`

MikeBLambert:
> @MikeBLambert I think I understand the issue. We're cleaning up the user input to use to compare, but what's actually being added to the db isn't "cleaned up" so it's not catching it. From research it seems that firebase is case sensitive, so if we also send a lowerCaseItemName as part of our formData and then compare that to cleanInput in the query we'd actually be comparing "apples to apples". 🍎🍏
> 
> ```
>  const snapshot = await queryCollection
>       .where('lowerCaseItemName', '==', cleanInput)
>       .get();
> ```
> 
> `db.collection(token).add({ ...formData, lowerCaseItemName: cleanInput });`
> 
> Another way I was thinking would be to get the user's list (without a query) and then filter it, but my concern is memory and performance with having to first get all the data and then search through it.

MikeBLambert:
@rachelrohrbach Of course! That makes total sense and now I understand how it was working for you and not for me. I think those are both totally viable options. I really like the idea of saving a query-able value as well as what is more or less a label for that value. 

Now, to go off on a slight tangent... if you think about it, this is separation of value and label is actually a pretty powerful pattern that you see a lot in software development. For example, let's look at the below <option /> tag, that's used inside of a <select /> tag (example text taken from w3schools):

```html
<select name="cars" id="cars">
  <option value="volvo">Volvo</option>
  <option value="saab">Saab</option>
  <option value="mercedes">Mercedes</option>
  <option value="audi">Audi</option>
</select>
```

Here, we give the option a `value` and then we render a label inside of each option tag. This allows us to separate out what you want something to display as from what its official value that we can reliably trust is. Your awesome suggestion is really the same thing :-)

src/components/AddItem.js
```diff
@@ -24,8 +26,24 @@ const AddItem = ({ token }) => {
         });
   };
 
-  const handleSubmit = (event) => {
+  const handleSubmit = async (event) => {
     event.preventDefault();
+    const itemNameInput = formData.itemName;
+
+    // check to see if the itemName already exists in the user's list - convert input to lowercase and remove any punctuation & white space
```
DeeDowns, line 33: I appreciate the comments explaining what is going on here

src/components/AddItem.js
```diff
@@ -46,6 +64,7 @@ const AddItem = ({ token }) => {
             value={formData.itemName}
             onChange={handleFormChange}
           />
+          {hidden ? <Duplicate /> : ''}
```
DeeDowns, line 67: Would this be the same as `{hidden ? <Duplicate /> : null}` ? I've never seen it done with an empty string, just wondering if there is an advantage to one over the other.

DeeDowns (approved):
I dont have much to add, the code looks clean, straightforward, and easy to read. I do appreciated the comments to explain what the

src/components/AddItem.js
veeteeran: The absence of a component using `null` makes much more sense than having an empty string. Thanks for pointing this out!

LINDGRENBA: I did some research and it looks like they are mostly interchangeable. They are both read as `falsy` by the JSX. However, I seem to remember hearing that there may be some unique cases where one might be used over the other, but I can't track down where I heard / read that.

src/components/Duplicate.js
```diff
@@ -0,0 +1,13 @@
...
+const Duplicate = () => {
+  return (
+    <div>
+      <p style={{ color: 'red' }}>
+        ❌ <b>That item is already in your list</b>
+      </p>
+    </div>
+  );
+};
+
```
LINDGRENBA, line 12: I really like the way you approached this AC. If there were additional places in the code that required error messages, I think a small refactor would make this reusable for multiple situations.

src/components/AddItem.js
```diff
@@ -24,8 +26,24 @@ const AddItem = ({ token }) => {
...
   };
 
-  const handleSubmit = (event) => {
+  const handleSubmit = async (event) => {
     event.preventDefault();
+    const itemNameInput = formData.itemName;
+
+    // check to see if the itemName already exists in the user's list - convert input to lowercase and remove any punctuation & white space
+    const cleanInput = itemNameInput.toLowerCase().replace(/[^\w\s]|/g, '');
+
```
LINDGRENBA, line 35: This may be a holdover from @MikeBLambert 's previous comment and you may have already addressed this, but just in case, when I viewed the application with Netlify and entered similar items with slight changes, I was able to add them to the list. Just wanted to pop this in here in case it's a tenacious bug.  
snapshot of list

LINDGRENBA (approved):
Overall, it looks great. Nice updates and I like how reusable the `Duplicate` error component could be in the future! Great job @veeteeran and @rachelrohrbach

src/components/AddItem.js
veeteeran: I think @rachelrohrbach and I fixed this issue. We also refactored the regex.

src/components/AddItem.js
```diff
@@ -46,6 +67,9 @@ const AddItem = ({ token }) => {
             value={formData.itemName}
             onChange={handleFormChange}
           />
+          {hidden ? (
```
laurenmbeatty, line 70: This is an unintuitive name to me. When I read this it says:

when hidden is true, then show this message. Maybe `hasDuplicate` and `setHasDuplicate`? something like that. That way it would be more like:

If this thing has a duplicate, then show this error message.

src/components/AddItem.js
```diff
@@ -24,9 +27,27 @@ const AddItem = ({ token }) => {
         });
   };
 
-  const handleSubmit = (event) => {
+  const handleSubmit = async (event) => {
     event.preventDefault();
```
laurenmbeatty, line 31: maybe hide the error message at the beginning of the submit. That way if the error is already shown, it will hide on submit, so we know we're getting a new error message.

src/components/AddItem.js
```diff
@@ -24,9 +27,27 @@ const AddItem = ({ token }) => {
...
+
+    // query the collection and filter for match - if match exists send the user an error message "this item already exists in your list..." & don't allow them to submit it
+    const queryCollection = db.collection(token);
+    const snapshot = await queryCollection
+      .where('cleanedUpItemName', '==', cleanInput)
+      .get();
+    if (!snapshot.empty) {
+      setHidden(true);
+      return;
+    }
```
laurenmbeatty, line 45: This totally works...nice!  

Another way to approach it would be to use the results that you already available to you (without making a new call to the db...would need to be passed down via context or props), and then either `filter` or `find` if there is a duplicate item. You would need to do the same `.toLowerCase().replace(/[\W_]*|/g, '')` on the `itemName` of the result item to compare. 

If you all end up re-factoring to pass down state via context or props, this would be possible.

laurenmbeatty (approved):
Nice job, y'all! This thing is really coming together.

Going forward, I would recommend _not_ pulling the other pair's branch into yours. It makes it hard to see which changes are yours and which are the other pair's. It's not a typical practice, to my knowledge.

src/components/AddItem.js
laurenmbeatty: Yes... 💯 %

src/components/AddItem.js
```diff
@@ -1,16 +1,19 @@
...
   itemName: '',
+  cleanedUpItemName: '',
   timeFrame: 7,
   lastPurchased: null,
 };
 
 const AddItem = ({ token }) => {
   const history = useHistory();
   const [formData, setFormData] = useState(initialFormState);
+  const [hidden, setHidden] = useState(false);
```
chrisdopuch, line 16: Reading through this code as someone who is unfamiliar, it's not quite clear to me what `hidden` is going to be used for. Maybe it hides the form, or parts of it? Maybe something else?

I would recommend a variable name here that makes it super clear, such as `shouldShowError` or something like that. I always like to prefix my boolean values using `is` or `should` or something to indicate that it's a boolean flag.

chrisdopuch: Oh hehe looks like @laurenmbeatty had the same suggestion here! https://github.com/the-collab-lab/tcl-17-smart-shopping-list/pull/23/files#r561408192

chrisdopuch (approved):
The demo worked great on netlify 👍 
It looks like it satisfied all the AC and I couldn't break it 😄 
Nice work! 👏

src/components/Error.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+const Error = ({ errorMessage }) => {
+  return (
+    <div>
+      <p style={{ color: 'red' }}>
+        <span role="img" aria-label="red X">
+          ❌
```
MikeBLambert, line 8: Love the X emoji!

src/components/AddItem.js
```diff
@@ -1,16 +1,19 @@
 import React, { useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { db } from '../lib/firebase';
+import Error from './Error';
 
 const initialFormState = {
   itemName: '',
+  cleanedUpItemName: '',
```
MikeBLambert, line 8: I don't think that we are using `cleanedUpItemName` from state anywhere, so we can remove from the initial state. It's also state that can be derive from other state, so we wouldn't really need it. :-)

MikeBLambert (approved):
It's working great for me. Nice job hunting down that bug and coming up with a great solution!

rachelrohrbach:
> Nice job, y'all! This thing is really coming together.
> 
> Going forward, I would recommend _not_ pulling the other pair's branch into yours. It makes it hard to see which changes are yours and which are the other pair's. It's not a typical practice, to my knowledge.

@laurenmbeatty yeah I did not intend to merge their branch 🤦

src/components/AddItem.js
```diff
@@ -24,9 +27,27 @@ const AddItem = ({ token }) => {
...
+      .where('cleanedUpItemName', '==', cleanInput)
+      .get();
+    if (!snapshot.empty) {
+      setHidden(true);
+      return;
+    }
+    // else add the user's inputted item to the database
+    db.collection(token).add({
+      ...formData,
+      cleanedUpItemName: cleanInput,
```
MikeBLambert, line 49: Nit - One alternative I had mentioned in Slack and idea for a future refactor (there are always alternatives and that's what makes things fun!) would be to make the `cleanInput` be the document's id rather than a property on the document itself. That could make retrieval of the document a little cleaner. If you did that, you may consider renaming `cleanInput` to something like `itemId`.
This 100% works as is and I think either way works great.

laurenmbeatty:
> > Nice job, y'all! This thing is really coming together.
> > Going forward, I would recommend _not_ pulling the other pair's branch into yours. It makes it hard to see which changes are yours and which are the other pair's. It's not a typical practice, to my knowledge.
> 
> @laurenmbeatty yeah I did not intend to merge their branch 🤦

🤗  I have been in many a git mess, for sure.

PR #25, opened 2021-01-21
change state variables names to be more semantic; include error state…

DeeDowns:
… change in onchange handler

## Description
This PR will allow users to share tokens and collaborate on other lists. A user can enter an existing token which will be saved to `localStorage`, the corresponding list will be collected from `Firestore` and the user will be redirected to the `list` view. If the list does not exist, the user will receive an error message and will not be redirected. 

## Related Issue
[#6](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/6)

## Acceptance Criteria
- [x] When the user does not already have a token in localStorage, on the onboarding/home screen, a simple form is displayed that allows the user to enter a token
- [x] Entering the token and hitting submit saves the token to localStorage, effectively giving them joint control of the list
- [x] On submit, show an error if the token does not exist
- [x] If they get an error message, allow them to try again or create a new list

## Type of Changes

### Before
welcome-before

### After
welcome-after
welcome-error

* One way to test the site is to click on the `details` button below to open the Netlify preview.

* Otherwise from your terminal, pull down this branch with git pull origin `BL-DD-join-existing-list` and check that branch out with git checkout `BL-DD-join-existing-list`
* Use `npm start` to launch the app in the browser.

* If you already have a token saved in `localStorage` you will need to clear the token first in order to test this feature. In your developer tools in the browser, navigate to `application`, select `local storage` and hit the `clear` icon to remove any pre-existing tokens from your local storage.

1. First enter `not real token` into the `join an existing list` field to get an error message for a token that doesn't exist in the database.
2. Then enter `chick gould nibs` to join an existing list that does already exist in the database.

### DIFF
--- a/src/components/TokenForm.js
+++ b/src/components/TokenForm.js
@@ -4,7 +4,7 @@ import Error from './Error';
 
 const TokenForm = ({ setToken }) => {
   const [tokenInput, setTokenInput] = useState('');
-  const [hidden, setHidden] = useState(true);
+  const [error, setError] = useState(false);
 
   const handleFormSubmit = async (event) => {
     event.preventDefault();
@@ -13,10 +13,10 @@ const TokenForm = ({ setToken }) => {
       .where('tokenName', '==', tokenInput)
       .get();
     if (tokenWeAreCheckingFor.empty) {
-      setHidden(false);
+      setError(true);
     } else {
       localStorage.setItem('userToken', tokenInput);
-      setHidden(true);
+      setError(false);
       setToken(tokenInput);
     }
   };
@@ -35,15 +35,14 @@ const TokenForm = ({ setToken }) => {
               value={tokenInput}
               onChange={(event) => {
                 setTokenInput(event.target.value);
+                setError(false);
               }}
               required
             />
           </label>
           <button type="submit">Join an existing list</button>
         </form>
-        {hidden ? (
-          ''
-        ) : (
+        {error && (
           <Error errorMessage="Hmmm... we couldn't find that list. Please try again or create a new list." />
         )}
       </div>


src/components/TokenForm.js
```diff
@@ -4,7 +4,7 @@ import Error from './Error';
 
 const TokenForm = ({ setToken }) => {
   const [tokenInput, setTokenInput] = useState('');
-  const [hidden, setHidden] = useState(true);
+  const [error, setError] = useState(false);
```
DeeDowns, line 7: Pretty much renaming the error state variable to be more semantic

src/components/TokenForm.js
```diff
@@ -35,15 +35,14 @@ const TokenForm = ({ setToken }) => {
               value={tokenInput}
               onChange={(event) => {
                 setTokenInput(event.target.value);
+                setError(false);
```
DeeDowns, line 38: Took suggestion from Chris to set the error state to false so that the error goes away while a user is typing

src/components/TokenForm.js
```diff
@@ -35,15 +35,14 @@ const TokenForm = ({ setToken }) => {
...
               }}
               required
             />
           </label>
           <button type="submit">Join an existing list</button>
         </form>
-        {hidden ? (
-          ''
-        ) : (
+        {error && (
```
DeeDowns, line 45: Chris also suggested to use an inline boolean guard to do conditional rendering instead of using a ternary

LINDGRENBA (approved):
Looks great!

MikeBLambert (approved):
Changes look straightforward and well implemented. 👍

PR #26, opened 2021-01-22
Add files from BL-DD-join-existing-list

rachelrohrbach:
## Description

This should add back the changes that I reverted when I did the git merge revert in my branch after accidentally merging their branch into ours. I took the latest versions of these files from BL-DD-join-existing-list branch.  

## Type of Changes

|  ✓  | :bug: Bug fix              |


LINDGRENBA:
Okay, awesome! We should probably merge this first. Then hopefully the message on the other PR about merge conflicts will resolve. Can I go ahead?

PR #27, opened 2021-01-23
Rr vt prevent duplicates

### LINKED ISSUE #7: 6. As a user, I want to be alerted when I’m entering an item that is the same as something already on my list so I can avoid duplicate items.

veeteeran:
## Description

Renamed `error` variables to be more explicit. Deleted error message when user enters input. Added `propTypes` to `Error` component.

## Related Issue
Closes #7

## Acceptance Criteria

- [x] Show an error message if the user tries to submit a new item that has the exact same name as an existing item
- [x] Show an error message if the user tries to submit a new item that has the same name as an existing item, where capitalization has been normalized and punctuation has been removed
- [x] The user’s original input should be what gets saved in the database

## Type of Changes

* Otherwise from your terminal, pull down this branch with `git pull origin RR-VT-prevent-duplicates` and check that branch out with `git checkout RR-VT-prevent-duplicates`
* From add item page enter an item that already exists to see the error message
* List view reflects exact user input

### DIFF
--- a/src/components/AddItem.js
+++ b/src/components/AddItem.js
@@ -32,7 +32,7 @@ const AddItem = ({ token }) => {
     const itemNameInput = formData.itemName;
 
     // check to see if the itemName already exists in the user's list - convert input to lowercase and remove any punctuation & white space
-    const cleanInput = itemNameInput.toLowerCase().replace(/^[\w\s]|/g, '');
+    const cleanInput = itemNameInput.toLowerCase().replace(/[\W_]*|/g, '');
 
     // query the collection and filter for match - if match exists send the user an error message "this item already exists in your list..." & don't allow them to submit it
     const queryCollection = db.collection(token);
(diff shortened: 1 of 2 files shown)


src/components/AddItem.js
```diff
@@ -10,12 +10,13 @@ const initialFormState = {
   lastPurchased: null,
 };
 
-const AddItem = ({ token }) => {
+const AddItem = ({ token, props }) => {
```
chrisdopuch, line 13: @veeteeran is `props` still being used here? 
(by the way, this object you are destructuring _is_ props, this is like saying `props.props`, just FYI)

chrisdopuch (approved):
Nice improvement, thanks for following up!
It worked great on the demo environment. 
My only comment was the unused `props` variable, but otherwise looks excellent! 🎸

src/components/AddItem.js
veeteeran: We forgot to delete props. Definitely not supposed to be in there. Thanks Chris!

MikeBLambert (approved):
👍

PR #28, opened 2021-01-26
Add user prompt to empty list view

LINDGRENBA:
## Description
This PR will add a prompt for the user to add a first `list item` if none currently exist. After clicking the prompt (eg: `button`) the user will be redirected to the Add Item page. This prompt helps guide the user and indicates which type of action they can take. 

## Related Issue
[#8](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/8)

## Acceptance Criteria
- [ ] The list view, when there are no items to display, should show a prompt (e.g., a button) for the user to add their first item
- [ ] When the user clicks on the add item Button, they should be routed to the Add Item page.

## Type of Changes

### Before
empty-list-view

### After
empty-list-view-updated

1.  One way to test the site is to click on the `details` button below to open the `Netlify preview`.
2. Otherwise from your terminal, pull down this branch with `git pull origin BL-RR-add-first-item` and check that branch out with `git checkout origin BL-RR-add-first-item` 
* Type `npm start` into your terminal to launch the app in your browser.
* You will need to clear any existing tokens from your `Local Storage` before you can test this feature. Open the  developer tools in your browser, then go to `Application` --> `Local Storage` and delete any existing tokens.
local-storage

* Click the `Create a New List` button to generate a token and go to `List` view. Then click the `Add an Item` button to be redirected to the `Add Item` form.

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -12,3 +12,7 @@ const List = ({ token }) => {
 };
 
 export default List;
+
+//Issue 7 AC:
+// The list view, when there are no items to display, should show a prompt (e.g., a button) for the user to add their first item
+// When the user clicks on the add item Button, they should be routed to the Add Item page.
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -14,7 +14,7 @@ const List = ({ token }) => {
         </section>
       )}
       <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
-      {loading && <h1>Loading...</h1>}
+      {loading && <h3>Loading...</h3>}
     </div>
   );
 };
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -2,7 +2,7 @@ import React from 'react';
 import useFirestore from '../hooks/useFirestore';
 
 const List = ({ token }) => {
-  const { docs } = useFirestore(token);
+  const { docs, loading } = useFirestore(token);
 
   return (
     <div>
@@ -13,13 +13,8 @@ const List = ({ token }) => {
           <a href="/add-item">Add an Item</a>
         </section>
       )}
-      <ul>
-        {docs ? (
-          docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)
-        ) : (
-          <li>Loading...</li>
-        )}
-      </ul>
+      <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
+      {loading && <h1>Loading...</h1>}
     </div>
   );
 };
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -13,7 +13,13 @@ const List = ({ token }) => {
           <a href="/add-item">Add an Item</a>
         </section>
       )}
-      <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
+      <ul>
+        {docs ? (
+          docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)
+        ) : (
+          <li>Loading...</li>
+        )}
+      </ul>
     </div>
   );
 };
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -2,7 +2,7 @@ import React from 'react';
 import useFirestore from '../hooks/useFirestore';
 
 const List = ({ token }) => {
-  const { docs, loading } = useFirestore(token);
+  const { docs } = useFirestore(token);
 
   return (
     <div>
@@ -14,7 +14,6 @@ const List = ({ token }) => {
         </section>
       )}
       <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
-      {loading && <h3>Loading...</h3>}
     </div>
   );
 };
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import useFirestore from '../hooks/useFirestore';
+import Error from './Error';
 
 const List = ({ token }) => {
-  const { docs } = useFirestore(token);
+  const { docs, errorMessage } = useFirestore(token);
 
   return (
     <div>
@@ -13,6 +14,7 @@ const List = ({ token }) => {
           <a href="/add-item">Add an Item</a>
         </section>
       )}
+      {errorMessage && <Error errorMessage={errorMessage} />}
       <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
     </div>
   );
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,11 +1,18 @@
 import React from 'react';
+import { useHistory } from 'react-router-dom';
 import useFirestore from '../hooks/useFirestore';
 
 const List = ({ token }) => {
   const { docs } = useFirestore(token);
+  const history = useHistory();
+
   return (
     <div>
       <h1>List</h1>
+      {/* check if list is empty - if empty display propmt, else show user's list */}
+      {docs.length === 0 && (
+        <button onClick={() => history.push('/add-item')}>Add an Item</button>
+      )}
       <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
     </div>
   );
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,16 +1,17 @@
 import React from 'react';
-import { useHistory } from 'react-router-dom';
 import useFirestore from '../hooks/useFirestore';
 
 const List = ({ token }) => {
   const { docs } = useFirestore(token);
-  const history = useHistory();
 
   return (
     <div>
       <h1>List</h1>
       {docs.length === 0 && (
-        <button onClick={() => history.push('/add-item')}>Add an Item</button>
+        <div>
+          <p>Your shopping list is currently empty.</p>
+          <a href="/add-item">Add an Item</a>
+        </div>
       )}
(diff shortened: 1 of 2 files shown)


rachelrohrbach:
We have the functionality working, button displays if user has an empty list and clicking on the button redirects the user to the Add Item page so they can start building a list.  We still need to add some verbiage to the List page (i.e. "Your shopping list is currently empty.").

src/components/List.js
```diff
@@ -1,14 +1,25 @@
...
+      {docs.length === 0 && (
+        <button onClick={() => history.push('/add-item')}>Add an Item</button>
+      )}
       <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
     </div>
   );
 };
 
 export default List;
+
```
LINDGRENBA, line 22: We decided not to add in any pop up message or modal as that seemed like more of a fun styling thing that we could do when it's time to style the app, if the rest of the team decides that would be a good idea. Thought - since we have `docs.length === 0` twice, can we refactor and simplify?

src/components/List.js
```diff
@@ -1,14 +1,27 @@
...
+//Issue 7 AC:
+// The list view, when there are no items to display, should show a prompt (e.g., a button) for the user to add their first item
```
MikeBLambert, line 26: Should we remove these comments?

MikeBLambert (approved):
Just tested it out and it works great. Hits the acceptance criteria for sure! :-)

One small nit-picky thing that I noticed when testing it is that if you have items in your shopping list, it displays the newly added button and message for a split second while loading the items. If you have the time and are looking to add some fun polish, you could look into adding a loading message / loading spinner while we are loading the data.

src/components/List.js
LINDGRENBA: Yes, we should def remove those. And I like the idea of a loader. That would be a nice solution to that little bug.

src/components/List.js
```diff
@@ -1,14 +1,27 @@
...
 
 const List = ({ token }) => {
   const { docs } = useFirestore(token);
+  const history = useHistory();
+
   return (
     <div>
       <h1>List</h1>
+      {docs.length === 0 && (
+        <div>
```
laurenmbeatty, line 13: If you wanted, you could use a `section` tag instead of a div, or just use a React.Fragment. Might make sense to have a container around the paragraph and button, though.

laurenmbeatty (approved):
Looks great y'all! Nice and simple. I like the use of `useHistory`. 💯

DeeDowns (approved):
Code looks clean, easy to read and straight forward. I like the decision to hold off from implementing a modal pop up until we decide how we are going to style the app later.

src/components/List.js
```diff
@@ -1,14 +1,27 @@
...
+          <p>Your shopping list is currently empty.</p>
+          <button onClick={() => history.push('/add-item')}>Add an Item</button>
```
chrisdopuch, line 15: From an accessibility perspective, you might wish to consider using an anchor tag here instead of a button. 
Why is that? Basically, it comes down to HTML semantics, and how screen-readers treat buttons vs links.
In general, if something is a link to another page or route, or to an external site, you should use an anchor.
If something controls a form, opens or closes a dialog, or does anything else dynamic inside the page, use a button.
Check out this article for more in-depth discussion: https://a11y-101.com/design/button-vs-link

That being said, it's not a trivial amount of work to style a link to _look_ like a button. I'd say this is an optional suggestion, though so feel free to skip if you want. It does work as-is just fine.

src/components/List.js
chrisdopuch: +1 to the idea of a loader! 😄  The flash of "you have no items in the list" while the data is loading could be confusing or jarring for our hypothetical users.

chrisdopuch (approved):
Nice work y'all, I tested the demo and it's working great 👍

src/components/List.js
rachelrohrbach: Thanks for sharing this information, Chris! Good to know 🙂

veeteeran (approved):
Hey all. The changes look great and working as expected on my end. Well written concise code. Chris' comment on accessibility and Lauren's about using the section tag is interesting. I have a basic understanding of these semantic tags and am still trying to figure out how to structure my HTML. These are a few Chrome extensions I've used for an accessibility project.

[axe](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd) for web accessibility testing

For a visual representation of heading [headingsMap](https://chrome.google.com/webstore/detail/headingsmap/flbjommegcjonpdmenkdiocclhjacmbi?hl=en)

[Landmarks](https://chrome.google.com/webstore/detail/landmark-navigation-via-k/ddpokpbjopmeeiiolheejjpkonlkklgp) to visually locate the landmarks on your pages

LINDGRENBA:
This is great. We have a lot of good updates we can make based on everyone's comments! Thank you. And thanks @veeteeran for the accessibility resources!

src/components/List.js
LINDGRENBA: Rachel and I updated the button to a link, but it sounds like the article is advising against styling a link to look like a button.

src/components/List.js
```diff
@@ -3,10 +3,23 @@ import useFirestore from '../hooks/useFirestore';
...
     <div>
       <h1>List</h1>
-      <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
+      {docs.length === 0 && (
+        <section>
+          <p>Your shopping list is currently empty.</p>
+          <a href="/add-item">Add an Item</a>
+        </section>
+      )}
+      <ul>
```
LINDGRENBA, line 16: Rachel and I tried two different ways of adding the `Loading...` UI, here is the other:
```
 {docs ? <ul>{docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul> : <div>Loading...</div>}
```
are there any opinions on which one might be a better approach? Also, the docs load so quickly for us that it was difficult to confirm that the `Loading...` is what shows up instead of the `"your shopping list is empty"` message. Are there any suggestions on how to confirm the Loading message is the actual element that's appearing.

MikeBLambert: With regards to checking whether this is working, you can try throttling the browser. Go to the network tab and click on the dropdown that says "online" 
image

I believe that the loading message may actually not be showing at all right now because we are checking for the truthy/falsy-ness of docs, which is always an array is always truthy.

What I think we'd want to do is add a loading state to our `useFirestore` hook, which we would set to true just before we call the firbase db and set to false just after. Then, we can have `useFirestore` return that loading state in the object that is currently returning docs. 

This would actually make our hook interface act a lot like really popular data fetching hooks like react-query (https://react-query.tanstack.com) and apollo client's useQuery hook (https://www.apollographql.com/docs/react/data/queries/). :-)

MikeBLambert: One more thing to consider, is that if we're monkeying around with the useFirestore hook, it may be a good idea to add a try/catch block around the call to the database and think about how we'd like the app to behave should there be an error returned from firebase. 

Any time we make an API call/are relying on an external dependency, it's always a good idea to think about what would happen if our "happy path" got a little less happy and the external dependency started acting up. Maybe we could do something like show a snackbar message (https://www.npmjs.com/package/react-simple-snackbar).

LINDGRENBA: Okay! So Rachel and I worked on this a bit. We tried adding a try catch and an error. Here is our `try catch` code in the `useFirestore` hook : 

useFirestore with try catch

And the code in the `List` component : 

List with error and loading

The issue we encountered : both the `your list is empty` and the `loading` message would appear at the same time after adding an item.

Without the `try-catch` we no longer have that issue, but now when you come to your list, even if it 
has grocery items, you'll see the `your list is empty` message until the docs load. Here's the `useFirestore` hook for that : 

useFirestore no try catch

We like the idea of using a snackbar message or alert message, but wanted to check with @DeeDowns and @veeteeran to see what they think.

src/hooks/useFirestore.js
```diff
@@ -3,21 +3,24 @@ import { db } from '../lib/firebase';
...
   const [docs, setDocs] = useState([]);
+  const [loading, setLoading] = useState(false);
 
   useEffect(() => {
     const unsubscribe = db.collection(collection).onSnapshot((snapshot) => {
       let documents = [];
       snapshot.forEach((doc) => {
         documents.push({ ...doc.data(), id: doc.id });
       });
+      setLoading(true);
```
MikeBLambert, line 14: I think we may want to throw this initial `setLoading` to be the first line inside the `useEffect` callback. (i.e. 
``useEffect(() => {
   setLoading(true);
  const unsubscribe = ...
}, [collection])

src/components/List.js
MikeBLambert: I believe that the issue may have to do with trying to perform the cleanup from inside of the `try` block. You may want to try either reassigning the `unsubscribe` that is declared outside of the `try` block or you could extract the data fetching into another function and return `unsubscribe`. I'll post examples in a sec, but I have to jump offline for a little bit.

MikeBLambert: Example returning unsubscribe from a callback: 
image

Example with reassignment of unsubscribe:
image

MikeBLambert (approved):
AC was met long ago and adding the loading state is above and beyond. :-) Nice job on this and way to go the extra mile!

src/components/List.js
LINDGRENBA: Thanks for the suggestions @MikeBLambert . We tried both and unfortunately the issue is stubborn and neither resolved the problem, so we focused on just adding the `try catch`
 and made a comment to help remind us to tackle the loading issue at a later point in time.

Here is an image of the error handling we added : 

error UI for list view

laurenmbeatty: That is my favorite error message ever. 🤣

PR #29, opened 2021-01-27
create checkbox for items, work on updating date

### LINKED ISSUE #9: 8. As a user, I want to mark an item on my shopping list as purchased so the app can learn how often I buy different items.
AC:

* User is able to tap a checkbox or similar UI element to mark an item in the list as purchased
* Item should be shown as checked for 24 hours after the purchase is made (i.e. we assume the user does not need to buy the item again for at least 1 day)

veeteeran:
## Description

When user taps on checkbox next to item on the list a `Date` object is created and the `lastPurchased` property in Firebase is updated. Once 24 hours have passed the boxes are unchecked.

## Related Issue

Closes #9

## Acceptance Criteria

- [x] User is able to tap a checkbox or similar UI element to mark an item in the list as purchased
- [x] Item should be shown as checked for 24 hours after the purchase is made (i.e. we assume the user does not need to buy the item again for at least 1 day)

## Type of Changes

### Before
Screen Shot 2021-01-27 at 10 31 08 AM

### After

**User taps to check items**

Screen Shot 2021-01-29 at 5 09 23 PM

**After 24 hours have passed boxes are unchecked**

Screen Shot 2021-01-29 at 5 12 09 PM

* Otherwise from your terminal, pull down this branch with `git pull origin DD-VT-check-item-on-list` and check that branch out with `git checkout DD-VT-check-item-on-list`
* From add item page click item on list

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -9,13 +9,24 @@ const List = ({ token }) => {
     if (purchasedDate === null) {
       return false;
     }
-    const today = new Date();
-    const daySincePurchased = purchasedDate.toDate();
+    // changed to using Date.now(), time given in milliseconds
+    const today = Date.now();
+    /* Passing today as milliseconds to new Date(), gives us a readable date format
+       Console logging so we can see this is the case */
+    console.log(new Date(today));
+    // 24 hours in milliseconds
+    const addOneDay = 86400000;
+    // Set to 5 seconds after purchase for testing, should read purchasedDate + addOneDay
+    const oneDayAfterPurchase = purchasedDate + 5000;
 
-    daySincePurchased.setDate(daySincePurchased.getDate() + 1); // Add one day to the last purchased date
+    // The line below is no longer necessary
+    // daySincePurchased.setDate(daySincePurchased.getDate() + 1); // Add one day to the last purchased date
 
-    // If today > daySincePurchased at least 24 hours have passed, return false to uncheck box
-    if (today >= daySincePurchased) {
+    // If today > oneDayAfterPurchase at least 24 hours have passed, return false to uncheck box
+    /* I simplified the condition after Chris pointed out that both were Date objects
+    and we didn't need the toString. I then figured we didn't need Date.parse since
+    we're working with Date objects */
+    if (today >= oneDayAfterPurchase) {
       return false;
     }
 
@@ -23,7 +34,7 @@ const List = ({ token }) => {
   };
 
   const handleCheckbox = async (event) => {
-    const currentDate = new Date();
+    const currentDate = Date.now(); // changed to using Date.now(), time given in milliseconds
     const queryCollection = await db.collection(token).doc(event.target.id);
     queryCollection.update({
       lastPurchased: currentDate,
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -10,18 +10,16 @@ const List = ({ token }) => {
       return false;
     }
     const today = new Date();
-    let daySincePurchased = purchasedDate.toDate();
+    const daySincePurchased = purchasedDate.toDate();
 
     daySincePurchased.setDate(daySincePurchased.getDate() + 1); // Add one day to the last purchased date
 
     // If today > daySincePurchased at least 24 hours have passed, return false to uncheck box
-    if (
-      Date.parse(today.toString()) >= Date.parse(daySincePurchased.toString())
-    ) {
+    if (today >= daySincePurchased) {
       return false;
-    } else {
-      return true;
     }
+
+    return true;
   };
 
   const handleCheckbox = async (event) => {
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -7,15 +7,13 @@ const List = ({ token }) => {
   const handleCheckbox = async (event) => {
     const currentDate = new Date();
     const nextDay = new Date();
-    // localStorage.nextDay = nextDay.setDate(nextDay.getDate() + 1)
-    console.log(event.target);
-    const queryCollection = await db
-      .collection(token)
-      .doc(event.target.id)
-      .get();
-    // const snapshot = await queryCollection.get()
-    queryCollection.data();
-    console.log(queryCollection.data());
+    // nextDay.setDate(nextDay.getDate() + 1)
+    // console.log(event.target);
+    const queryCollection = await db.collection(token).doc(event.target.id);
+    queryCollection.update({
+      lastPurchased: currentDate,
+      nextDay: nextDay,
+    });
   };
   return (
     <div>
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,12 +1,39 @@
 import React from 'react';
 import useFirestore from '../hooks/useFirestore';
+import { db } from '../lib/firebase';
 
 const List = ({ token }) => {
   const { docs } = useFirestore(token);
+  const handleCheckbox = async (event) => {
+    const currentDate = new Date();
+    const nextDay = new Date();
+    // localStorage.nextDay = nextDay.setDate(nextDay.getDate() + 1)
+    console.log(event.target);
+    const queryCollection = await db
+      .collection(token)
+      .doc(event.target.id)
+      .get();
+    // const snapshot = await queryCollection.get()
+    queryCollection.data();
+    console.log(queryCollection.data());
+  };
   return (
     <div>
       <h1>List</h1>
-      <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
+      <ul style={{ listStyleType: 'none' }}>
+        {docs &&
+          docs.map((doc) => (
+            <li key={doc.id}>
+              <input
+                type="checkbox"
+                name={doc.itemName}
+                id={doc.id}
+                onClick={handleCheckbox}
+              />
+              {doc.itemName}
+            </li>
+          ))}
+      </ul>
     </div>
   );
 };
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,36 +1,42 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import useFirestore from '../hooks/useFirestore';
 import { db } from '../lib/firebase';
 
 const List = ({ token }) => {
   const { docs } = useFirestore(token);
-  const [checked, setChecked] = useState();
-  // console.log(checked)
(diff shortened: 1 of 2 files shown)


src/components/List.js
```diff
@@ -1,12 +1,56 @@
...
-      <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
+      <ul style={{ listStyleType: 'none' }}>
+        {docs &&
+          docs.map((doc) => (
+            <li key={doc.id}>
+              <input
+                type="checkbox"
+                name={doc.itemName}
+                id={doc.id}
+                onClick={handleCheckbox}
```
laurenmbeatty, line 46: Instead of `onClick`, this function should be `onChange`. 👍

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
   const { docs } = useFirestore(token);
+
+  const checkPurchasedDate = (purchasedDate) => {
+    if (purchasedDate === null) {
+      return false;
+    }
+    const today = new Date();
+    let daySincePurchased = purchasedDate.toDate();
+
+    daySincePurchased.setDate(daySincePurchased.getDate() + 1); // Add one day to the last purchased date
```
MikeBLambert, line 15: Thanks for the comments! :-)

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
-      <ul>{docs && docs.map((doc) => <li key={doc.id}>{doc.itemName}</li>)}</ul>
+      <ul style={{ listStyleType: 'none' }}>
+        {docs &&
+          docs.map((doc) => (
+            <li key={doc.id}>
+              <input
+                type="checkbox"
+                name={doc.itemName}
+                id={doc.id}
+                onChange={handleCheckbox}
```
chrisdopuch, line 45: Right now, if the user clicks on an already checked checkbox it will stay locked in the checked state. 
What if the user accidentally clicked the wrong one by mistake? Being able to clear the checkbox might be useful.

This wasn't explicitly called out in the acceptance criteria, so don't feel like you need to make this change. Just an idea worth thinking about!

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
+    ) {
+      return false;
+    } else {
+      return true;
+    }
+  };
+
+  const handleCheckbox = async (event) => {
+    const currentDate = new Date();
+    const queryCollection = await db.collection(token).doc(event.target.id);
```
chrisdopuch, line 29: Nice use of the id field on the event to track which item to update 👍

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
+      return false;
+    }
+    const today = new Date();
+    let daySincePurchased = purchasedDate.toDate();
+
+    daySincePurchased.setDate(daySincePurchased.getDate() + 1); // Add one day to the last purchased date
+
+    // If today > daySincePurchased at least 24 hours have passed, return false to uncheck box
+    if (
+      Date.parse(today.toString()) >= Date.parse(daySincePurchased.toString())
```
chrisdopuch, line 19: `today` and `daySincePurchased` are both of type `Date` right? Is it possible to remove the `.toString` and the `Date.parse` and just use them directly as Date objects? (I could be wrong, too! I'm not sure)

src/components/List.js
veeteeran: Thanks @chrisdopuch. We noticed that issue as well and that was brought up during office hours today. It sounds like something that will be handled down the road

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
 
 const List = ({ token }) => {
   const { docs } = useFirestore(token);
+
+  const checkPurchasedDate = (purchasedDate) => {
+    if (purchasedDate === null) {
+      return false;
+    }
+    const today = new Date();
+    let daySincePurchased = purchasedDate.toDate();
```
chrisdopuch, line 13: This variable could also be a `const` since it's never reassigned.

chrisdopuch (commented):
So far so good! I'll defer to the rest of the team for the initial review, just wanted to give you some early feedback. 
The demo worked great 👍 
I'll check on the 24 hour un-check tomorrow 😄

src/components/List.js
chrisdopuch: Sounds good to me, glad you are thinking about it too 🙌

src/components/List.js
veeteeran: Looking into this now. I think we had that in to console log something at some point. I've deleted and it still works. Going to do some more testing before I push the change 😄

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
+
+    daySincePurchased.setDate(daySincePurchased.getDate() + 1); // Add one day to the last purchased date
+
+    // If today > daySincePurchased at least 24 hours have passed, return false to uncheck box
+    if (
```
MikeBLambert, line 18: Nit-pick: 
Any time you find yourself with a conditional or ternary, in which you are returning just `true` or `false`, you can actually just return the conditional statement itself. If you're returning false if the conditional is true, as is the case here, then you can then negate the conditional statement and return it.

src/components/List.js
MikeBLambert, line 13: Really interesting approach here :-) I really like the idea of adding a day to the purchase day and then comparing that to today.

I think that the variable naming is a little confusing though. We're calling this variable `daysSincePurchased`, so I would expect it to be an integer indicating how many days have passed since it was purchased. However, it's actually a firebase timestamp of purchased date that is converted to a date object. Then, when we mutate it, it becomes that date plus one day.
I wonder if the team has any ideas for what would be a some good names here?

PS - one of my favorite programming quotes is: 

"There are only two hard things in Computer Science: cache invalidation and naming things."

-- Phil Karlton

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
+      Date.parse(today.toString()) >= Date.parse(daySincePurchased.toString())
+    ) {
+      return false;
+    } else {
+      return true;
+    }
+  };
+
+  const handleCheckbox = async (event) => {
+    const currentDate = new Date();
```
MikeBLambert, line 28: Non-rhetorical question for anyone and everyone - would it be preferable to save Date.now(), which would give us a unix timestamp, rather than `new Date()`, which appears to get converted into a Firebase timestamp (https://firebase.google.com/docs/reference/android/com/google/firebase/Timestamp)?
For me I feel like a unix timestamp would be more straightforward to deal with and leaves our logic slightly less tightly-coupled to firebase, but I don't know if there are advantages to doing it this away though that I'm not aware of.
If we were to use a Date.now(), in `checkPurchasedDate`, I believe that we would convert to a JS date with something that looks like `new Date(purchasedDate)`

src/components/List.js
MikeBLambert: I'll second that one! Nice one!

src/components/List.js
veeteeran: Hi Mike. I've simplified the code a bit from the comments Chris made above. When I have this `return (today > daySincePurchased);` I'm unable to check the boxes. Not sure why this is happening. Your suggestion makes sense to me

Without a `return true;` I get this error
Screen Shot 2021-01-29 at 6 10 03 PM

MikeBLambert: I believe that this error indicates that you are returning undefined somehow. when you console.log(today > daySincePurchased), what does it evaluate to? Are you returning early anywhere in the function?

MikeBLambert: I believe that you need to switch it to `daySincePurchased > today`.

veeteeran: So much fun! 🤯 I need to think on this before I push any changes 😃

src/components/List.js
veeteeran: @MikeBLambert dates are making my head hurt also 😅 I'm going to step away for the night and think some more. Thanks for all the  comments!

src/components/List.js
rachelrohrbach: Yes, the comments are great!

rachelrohrbach (approved):
This was a tough issue, so way to go! Dates are 🤯 I appreciate the comments in the code and it passes the AC!

src/components/List.js
LINDGRENBA: For ideas on variable names, maybe something regarding a timestamp, like timestampOfPurchasedDate or timestampOfCurrentDate (those are a little verbose, but something in that realm or ballpark maybe) ?

LINDGRENBA (approved):
This was a really tough one and I know both of you put in a lot of work this week. Great job on a very difficult issue. I think there were a lot of great questions and suggestions brought up by the rest of the team about things to consider, but it works well and you've met the AC! Awesome 😄 👍

chrisdopuch (approved):
This is looking great! I was able to get the checkboxes to uncheck after 24 hours. Nice work y'all! 👍 
(sorry I didn't get around to the review yesterday, my night got away from me!)

src/components/List.js
DeeDowns: I believe we wrote it that way because we were getting confused with the logic and wanted to make it easier for our brains to understand what was going on 😅

src/components/List.js
DeeDowns: I agree, the naming conventions are a bit confusing, I like your suggestion Brittany

src/components/List.js
```diff
@@ -1,12 +1,54 @@
...
+  const checkPurchasedDate = (purchasedDate) => {
+    if (purchasedDate === null) {
+      return false;
+    }
+    // changed to using Date.now(), time given in milliseconds
+    const today = Date.now();
+    // 24 hours in milliseconds
+    const addOneDay = 86400000;
+    // Add 24 hours to purchased
+    const oneDayAfterPurchase = purchasedDate + addOneDay;
```
MikeBLambert, line 17: This is so easy to read :-)

MikeBLambert (approved):
Looks good! Great job sticking to this one. Dates are frustrating, to say the least, and y'all did a wonderful job dealing with them! 

Note - prior to merging, you will want to resolve the merge conflicts.

LINDGRENBA:
Just in case anyone is wondering, Viet and Dee have already fixed the error handling by switching it from a hard coded message to an empty string as initial value. Rachel and I had thought we'd taken care of that, but somehow it snuck by us. Anyways, once this branch is merged with main, that should be resolved and work as expected.

PR #30, opened 2021-02-01
Add filtering of user's list

### LINKED ISSUE #10: 9. As a user, I want to filter my shopping list to make it easier to locate an item in the list.
AC:

* Display a text field above the top of the shopping list
* As the user types into the field, the list should narrow to display only items that contain the text the user entered in the filter field
* When the field has text in it, the user should be able to tap a UI element (e.g., with an "X" button next to the field) to clear the field
* The filter text should match any part of the item name (i.e. it should _not_ only match from the start of the string)

DeeDowns:
## Description
Adds a search bar to the top of the shopping list, with the ability to filter through the list items and an option to clear the search input

## Related Issue
closes #10 

## Acceptance Criteria

- [ ] Display a text field above the top of the shopping list

- [ ] As the user types into the field, the list should narrow to display only items that contain the text the user entered in the filter field

- [ ] When the field has text in it, the user should be able to tap a UI element (e.g., with an "X" button next to the field) to clear the field

- [ ] The filter text should match any part of the item name (i.e. it should not only match from the start of the string)

## Type of Changes

### Before
Screen Shot 2021-02-01 at 2 54 16 PM

### After

filteringScreenShot

filterListAnimatedGif

2. Otherwise from your terminal, pull down this branch with git pull origin RR-DD-filter-list and check that branch out with git checkout origin RR-DD-filter-list
   * On the list page start typing in the Filter Items input field and you will see the name of items that match what you have typed appear in the list, and to clear the search you can simply click the Reset button and the full list will display.

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -65,3 +65,11 @@ const List = ({ token }) => {
 };
 
 export default List;
+
+/*
+Display a text field above the top of the shopping list
+As the user types into the field, the list should narrow to display only items that contain the text the user entered in the filter field
+When the field has text in it, the user should be able to tap a UI element (e.g., with an "X" button next to the field) to clear the field
+The filter text should match any part of the item name (i.e. it should not only match from the start of the string)
+
+*/
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,10 +1,20 @@
-import React from 'react';
+import { doc } from 'prettier';
+import React, { useState } from 'react';
 import useFirestore from '../hooks/useFirestore';
 import { db } from '../lib/firebase';
 import Error from './Error';
 
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
+  const [searchInput, setSearchInput] = useState('');
+
+  const handleSearchChange = (event) => {
+    setSearchInput(event.target.value);
+  };
+
+  const handleClear = () => {
+    setSearchInput('');
+  };
 
   const checkPurchasedDate = (purchasedDate) => {
     if (purchasedDate === null) {
@@ -35,6 +45,19 @@ const List = ({ token }) => {
   return (
     <div>
       <h1>List</h1>
+      <form>
+        <label for="search-bar">Filter Items</label>
+        <br />
+        <input
+          type="text"
+          name="search-bar"
+          id="search-bar"
+          placeholder="Start typing here..."
+          value={searchInput}
+          onChange={handleSearchChange}
+        />
+        <input type="reset" onClick={handleClear} />
+      </form>
 
       {docs.length === 0 && (
         <section>
@@ -47,18 +70,22 @@ const List = ({ token }) => {
 
       <ul style={{ listStyleType: 'none' }}>
         {docs &&
-          docs.map((doc) => (
-            <li key={doc.id}>
-              <input
-                type="checkbox"
-                name={doc.itemName}
-                id={doc.id}
-                onChange={handleCheckbox}
-                checked={checkPurchasedDate(doc.lastPurchased)}
-              />
-              {doc.itemName}
-            </li>
-          ))}
+          docs
+            .filter((doc) =>
+              doc.itemName.toLowerCase().includes(searchInput.toLowerCase()),
+            )
+            .map((doc) => (
+              <li key={doc.id}>
+                <input
+                  type="checkbox"
+                  name={doc.itemName}
+                  id={doc.id}
+                  onChange={handleCheckbox}
+                  checked={checkPurchasedDate(doc.lastPurchased)}
+                />
+                {doc.itemName}
+              </li>
+            ))}
       </ul>
     </div>
   );
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -71,15 +71,16 @@ const List = ({ token }) => {
       <ul style={{ listStyleType: 'none' }}>
         {docs &&
           docs
-            .filter((doc) =>
-              doc.itemName
-                .toLowerCase()
-                .includes(searchInput.toLowerCase().trim()),
+            ?.filter((doc) =>
+              doc?.itemName
+                ?.toLowerCase()
+                ?.includes(searchInput.toLowerCase().trim()),
             )
-            .map((doc) => (
+            ?.map((doc) => (
               <li key={doc.id}>
                 <input
                   type="checkbox"
+                  aria-label="purchased-checkbox"
                   name={doc.itemName}
                   id={doc.id}
                   onChange={handleCheckbox}
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -92,11 +92,3 @@ const List = ({ token }) => {
 };
 
 export default List;
-
-/*
-Display a text field above the top of the shopping list
-As the user types into the field, the list should narrow to display only items that contain the text the user entered in the filter field
-When the field has text in it, the user should be able to tap a UI element (e.g., with an "X" button next to the field) to clear the field
-The filter text should match any part of the item name (i.e. it should not only match from the start of the string)
-
-*/
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -1,4 +1,3 @@
-import { doc } from 'prettier';
 import React, { useState } from 'react';
 import useFirestore from '../hooks/useFirestore';
 import { db } from '../lib/firebase';
@@ -45,25 +44,26 @@ const List = ({ token }) => {
   return (
     <div>
       <h1>List</h1>
-      <form>
-        <label for="search-bar">Filter Items</label>
-        <br />
-        <input
-          type="text"
-          name="search-bar"
-          id="search-bar"
-          placeholder="Start typing here..."
-          value={searchInput}
-          onChange={handleSearchChange}
-        />
-        <input type="reset" onClick={handleClear} />
-      </form>
 
-      {docs.length === 0 && (
+      {docs.length === 0 ? (
         <section>
           <p>Your shopping list is currently empty.</p>
           <a href="/add-item">Add an Item</a>
         </section>
+      ) : (
+        <form>
+          <label for="search-bar">Filter Items</label>
+          <br />
+          <input
+            type="text"
+            name="search-bar"
+            id="search-bar"
+            placeholder="Start typing here..."
+            value={searchInput}
+            onChange={handleSearchChange}
+          />
+          <input type="reset" onClick={handleClear} />
+        </form>
       )}
 
       {errorMessage && <Error errorMessage={errorMessage} />}
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -51,8 +51,8 @@ const List = ({ token }) => {
           <a href="/add-item">Add an Item</a>
         </section>
       ) : (
-        <form>
-          <label for="search-bar">Filter Items</label>
+        <div>
+          <label htmlFor="search-bar">Filter Items</label>
           <br />
           <input
             type="text"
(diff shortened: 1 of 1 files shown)


src/components/List.js
```diff
@@ -36,29 +45,47 @@ const List = ({ token }) => {
     <div>
       <h1>List</h1>
 
-      {docs.length === 0 && (
+      {docs.length === 0 ? (
```
rachelrohrbach, line 48: We changed this to a ternary so that you would either see the prompt to add your first item or the search bar for filtering if a list exists.

src/components/List.js
```diff
@@ -36,29 +45,47 @@ const List = ({ token }) => {
...
         <section>
           <p>Your shopping list is currently empty.</p>
           <a href="/add-item">Add an Item</a>
         </section>
+      ) : (
+        <form>
+          <label for="search-bar">Filter Items</label>
+          <br />
```
veeteeran, line 56: Nice use of `<br />` to separate the label and input

src/components/List.js
```diff
@@ -36,29 +45,47 @@ const List = ({ token }) => {
...
-                id={doc.id}
-                onChange={handleCheckbox}
-                checked={checkPurchasedDate(doc.lastPurchased)}
-              />
-              {doc.itemName}
-            </li>
-          ))}
+          docs
+            .filter((doc) =>
+              doc.itemName.toLowerCase().includes(searchInput.toLowerCase()),
```
veeteeran, line 75: Great idea to make the search not case sensitive! 🔤

veeteeran (commented):
Nice work you two! I also really liked having the gif in the PR to showing the search bar in action!

veeteeran (approved):

src/components/List.js
LINDGRENBA: Agreed! It's a nice straightforward solution.

LINDGRENBA (approved):
The search bar and filtering feels really smooth! I like the label 'reset' for the button that clears the text - it seems very semantically appropriate. All of the AC are met!  One thing I noticed (and since it's not part of the AC, maybe could be a separate issue) is that I expected the "Add Item" link to still be present and as a user it surprised me a little that it went away. I'm not sure we were thinking about this when we added it in last week. Should this be something we refactor at some point so that it's still present even when there are items in the list?

DeeDowns:
> The search bar and filtering feels really smooth! I like the label 'reset' for the button that clears the text - it seems very semantically appropriate. All of the AC are met! One thing I noticed (and since it's not part of the AC, maybe could be a separate issue) is that I expected the "Add Item" link to still be present and as a user it surprised me a little that it went away. I'm not sure we were thinking about this when we added it in last week. Should this be something we refactor at some point so that it's still present even when there are items in the list?

To me it seems more like the add item link goes hand in hand with the _Your shopping list is currently empty_ prompt when a list is empty. So when a list isn't empty the prompt to add an item (including the link) should no longer be there. The add item nav at the bottom of the page is still present for a user to add an item. This makes sense in my head from a user's standpoint, but I'm now interested in seeing what everyone else has to say 🤔  so thanks for bringing that up.

veeteeran:
I think I would keep things the way they are right now. Having the link go away in favor in the search bar is nice since we have the `Add Item` button at the bottom of the page. And when the list is empty having the option to tap the link or button feel natural to me.

LINDGRENBA:
I think what you're saying, @DeeDowns and @veeteeran, about the link makes sense to me 👍

rachelrohrbach:
I was thinking the same thing as @DeeDowns and @veeteeran since we already had the link and prompt disappearing when we displayed the user's list.

laurenmbeatty:
@rachelrohrbach and @DeeDowns what's a good token to use to test? Thx!

src/components/List.js
```diff
@@ -36,29 +45,47 @@ const List = ({ token }) => {
...
 
-      {docs.length === 0 && (
+      {docs.length === 0 ? (
         <section>
           <p>Your shopping list is currently empty.</p>
           <a href="/add-item">Add an Item</a>
         </section>
+      ) : (
+        <form>
+          <label for="search-bar">Filter Items</label>
```
laurenmbeatty, line 55: I think this needs to be `htmlFor`? `for` is reserved in JavaScript

src/components/List.js
```diff
@@ -36,29 +45,47 @@ const List = ({ token }) => {
...
+          <br />
+          <input
+            type="text"
+            name="search-bar"
+            id="search-bar"
+            placeholder="Start typing here..."
+            value={searchInput}
+            onChange={handleSearchChange}
+          />
+          <input type="reset" onClick={handleClear} />
```
laurenmbeatty, line 65: Oh! TIL about input `type="reset"` cool~

src/components/List.js
laurenmbeatty: Very nice! You might want to think about adding `.trim()`, which will trim the whitespace around either side of the value. That way if a user searches for ` apples` they will still find their `apples`.

laurenmbeatty:
Would someone mind testing if we need the `form` element in this case, since we're not submitting anything? Try removing the `form` just for a test, and then running a Lighthouse score in dev tools:
image

Right now I'm getting an accessibility score of 90, because our checkboxes don't have labels (we should make a ticket/issue for that), but if the score doesn't change we don't need it. I'm searching and I'm just not finding an answer.

It seems like having a form would be fine, but screen readers do interesting things when entering form mode, so I'm not 100% sure.

laurenmbeatty (approved):
LGTM 👍  Very clean solution. Just the one `htmlFor` change, think about `.trim()`, and maybe test if you need a form element? Nice work!

rachelrohrbach:
@laurenmbeatty I removed the `form` element and ran Lighthouse and still got an accessibility score of 90.

rachelrohrbach:
> Would someone mind testing if we need the `form` element in this case, since we're not submitting anything? Try removing the `form` just for a test, and then running a Lighthouse score in dev tools:
> image
> 
> Right now I'm getting an accessibility score of 90, because our checkboxes don't have labels (we should make a ticket/issue for that), but if the score doesn't change we don't need it. I'm searching and I'm just not finding an answer.
> 
> It seems like having a form would be fine, but screen readers do interesting things when entering form mode, so I'm not 100% sure.

I removed the `form` element and ran Lighthouse and still got an accessibility score of 90.

rachelrohrbach:
> @rachelrohrbach and @DeeDowns what's a good token to use to test? Thx!
`stroll ames knack` has a lot of items in the list

laurenmbeatty:
> I removed the `form` element and ran Lighthouse and still got an accessibility score of 90.

Seems like it could be removed then. 👍  Thanks for checking!

src/components/List.js
DeeDowns: You are correct, will make those changes.

src/components/List.js
DeeDowns: > Very nice! You might want to think about adding `.trim()`, which will trim the whitespace around either side of the value. That way if a user searches for ` apples` and accidentally types a space at the beginning or end, they will still find them.

So on the `seachInput.toLowerCase()`also chain on `.trim()`?

DeeDowns: Never mind on the question, added trim and it works 👍

src/components/List.js
```diff
@@ -36,29 +45,49 @@ const List = ({ token }) => {
...
-                type="checkbox"
-                name={doc.itemName}
-                id={doc.id}
-                onChange={handleCheckbox}
-                checked={checkPurchasedDate(doc.lastPurchased)}
-              />
-              {doc.itemName}
-            </li>
-          ))}
+          docs
```
MikeBLambert, line 73: One thing that it's always good to be vigilantly on guard for, especially for data coming from an API that we don't have much control over, is null-pointer errors. Any time that there's a bunch of chained dot-notation, there's always an increased risk of this. 

In the past, we would have to use phrases like `doc && doc.itemName && doc.itemName.toLowerCase()...` or rely on third party utilities like lodash get (https://lodash.com/docs/4.17.15#get) in order to mitigate this risk.

Luckily, the brand new JS feature of optional-chaining (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) allows us to super easily add null-pointer error protection into our code. Even more luckily, it appears our project appears to be set up to use this modern JS feature!

If we were to use optional chaining here, I think it would look something like:

```javascript
docs
          ?.filter((doc) =>
            doc?.itemName
              ?.toLowerCase()
              ?.includes(searchInput?.toLowerCase()?.trim?.()),
          )
          ?.map(...
```
(note, I don't use optional chaining much b/c my project at work doesn't support it, so definitely triple-check my suggestion)

src/components/List.js
MikeBLambert: Totally! I didn't know about `type="reset"`!

MikeBLambert (approved):
Great job! Code works great and is super easy to read. There are a few open suggestions in the comments, but they all look to be suggested improvements beyond the AC. :-)

src/components/List.js
rachelrohrbach: Thanks for sharing @MikeBLambert. 
I find it interesting that optional-chaining isn't supported by Samsung Internet which is the official browser for Samsung devices (though apparently they are only [2.52%](https://caniuse.com/?search=optional%20chaining%20operator%20%3F) of global users), so if we were releasing the app for widespread use that might be something we would want to consider.

src/components/List.js
```diff
@@ -36,29 +45,49 @@ const List = ({ token }) => {
...
 
-      {docs.length === 0 && (
+      {docs.length === 0 ? (
         <section>
           <p>Your shopping list is currently empty.</p>
           <a href="/add-item">Add an Item</a>
         </section>
+      ) : (
+        <div>
+          <label htmlFor="search-bar">Filter Items</label>
```
chrisdopuch, line 55: Way to add a label element for the input!

Did you know that the checkboxes in this form are missing their labels? I get a warning about that when I audit the page with the Axe accessibility extension. You can learn more about it here: https://www.deque.com/axe/

Totally optional, but if you wanted to improve the page's accessibility you could also add labels for those checkboxes!

chrisdopuch (approved):
Well done y'all! 🎉 
It worked great on netlify when I QA tested it, no problems there!
The code looks great too, I just had one optional comment about accessibility.

src/components/List.js
MikeBLambert: Absolutely, that is totally the kind of thing that you'd want to consider and I love that you brought that consideration up. In fact, I believe that long, long ago (or at least some time before the three years ago when I became a developer...) a lot of time and effort was spent getting code to work with different browsers.

Lucky for us though, our project uses Babel, which takes care of translating our fancy, modern JS into boring, old, "I remember when I used to walk to school in the snow, uphill, both ways..." JS.

If you open up the node_modules, you can see that there is a dependency called @babel/plugin-proposal-optional-chaining. This is the package that specifically deals with optional-chaining.
image

Now, if you look towards the bottom of our package.json, you'll see that there's a section called "browserlist". I believe that this is what Babel references when it does this translation. (I'm pretty fuzzy on a lot of this stuff though, so I wouldn't be surprised or insulted if I was corrected on the details...) Interestingly enough, in looking at [browserlist](https://www.npmjs.com/package/browserslist#browsers), it appears that they actually reference the site that you linked to in your reply. 🙌 🙌

rachelrohrbach: Oh right, how could I forget about the "magic" of Babel 🧙

PR #31, opened 2021-02-02
Add estimate functionality to handleCheck method.

veeteeran:
## Description
This PR updates the database with the `estimatedNextPurchase` date and `numberOfPurchases.` The function [/src/lib/estimates.js](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/blob/main/src/lib/estimates.js)  is used to generate the number of days until next purchase.

## Related Issue
Closes [#11](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/11)

## Acceptance Criteria
- [x] When a purchase is recorded, the estimated number of days until the next purchase date should be calculated and recorded in the database

## Type of Changes

| ✓ | :sparkles: New feature     |

## Updates
No UI updates for this feature

* Otherwise from your terminal, pull down this branch with `git pull origin BL-VT-estimate-next-purchase` and check that branch out with `git checkout BL-VT-estimate-next-purchase`
* For PR review `npm install --save date-fns`, after review finished `npm uninstall date-fns`
* Create a new list and add items or if a list exists click the List button
* Tap a check box of any item on the list
* Open Firebase and confirm the fields `estimatedNextPurchase` and `numberOfPurchases` are updated

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -3,7 +3,7 @@ import useFirestore from '../hooks/useFirestore';
 import { db } from '../lib/firebase';
 import Error from './Error';
 import calculateEstimate from './../lib/estimates';
-import { differenceInDays } from 'date-fns';
+import { differenceInDays, addDays } from 'date-fns';
 
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
@@ -69,9 +69,23 @@ const List = ({ token }) => {
 
     queryCollection.update({
       lastPurchased: currentDate,
-      daysUntilNextPurchase: daysUntilNextPurchase,
-      numberOfPurchases: numberOfPurchases,
+      timeFrame: daysUntilNextPurchase,
+      numberOfPurchases,
     });
+  };
+
+  /* Function updates daysUntilNextPurchase when user reloads the app
+   * Used below when rendering List
+   */
+  const getDaysUntilNextPurchase = (doc) => {
+    const nextPurchasedDate = addDays(doc.lastPurchased, doc.timeFrame);
+
+    /* Added 1 here since we're working with full days
+     * if user taps a checkbox and the timeFrame was 7 days, 6 would have been displayed
+     */
+    const daysUntilNextPurchase =
+      differenceInDays(nextPurchasedDate, Date.now()) + 1;
+    return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
   return (
@@ -89,26 +103,29 @@ const List = ({ token }) => {
 
       <ul style={{ listStyleType: 'none' }}>
         {docs &&
-          docs.map((doc) => (
-            <li key={doc.id}>
-              <input
-                type="checkbox"
-                name={doc.itemName}
-                id={doc.id}
-                onChange={handleCheckbox}
-                checked={checkPurchasedDate(doc.lastPurchased)}
-                //disabled={checkPurchasedDate(doc.lastPurchased)} // Commented out for PR purposes only
-              />
-              {doc.itemName}
-              {doc.numberOfPurchases > 0 ? (
-                <p>
-                  Time until next purchase: {doc.daysUntilNextPurchase} days
-                </p>
-              ) : (
-                <p>You haven't purchased {doc.itemName} yet.</p>
-              )}
-            </li>
-          ))}
+          docs.map((doc) => {
+            return (
+              <li key={doc.id}>
+                <input
+                  type="checkbox"
+                  name={doc.itemName}
+                  id={doc.id}
+                  onChange={handleCheckbox}
+                  checked={checkPurchasedDate(doc.lastPurchased)}
+                  //disabled={checkPurchasedDate(doc.lastPurchased)} // Commented out for PR purposes only
+                />
+                {doc.itemName}
+                {doc.numberOfPurchases > 0 ? (
+                  <p>
+                    Time until next purchase: {getDaysUntilNextPurchase(doc)}{' '}
+                    days
+                  </p>
+                ) : (
+                  <p>You haven't purchased {doc.itemName} yet.</p>
+                )}
+              </li>
+            );
+          })}
       </ul>
     </div>
   );
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -9,6 +9,7 @@ import {
   addMilliseconds,
   toDate,
   getTime,
+  differenceInDays,
 } from 'date-fns';
 
 const List = ({ token }) => {
@@ -39,15 +40,19 @@ const List = ({ token }) => {
     // listItem is our current snapshot of the doc
     const listItem = docs.filter((doc) => doc.id === queryCollection.id)[0];
     const previouslyPurchasedDate = listItem.lastPurchased;
-    const latestInterval = currentDate - previouslyPurchasedDate;
-
+    // latestInterval is the number of full day periods between two dates
+    const latestInterval = differenceInDays(
+      currentDate,
+      previouslyPurchasedDate,
+    );
     // If item has not been purchased set to 1, else increment by 1
     const numberOfPurchases = (listItem.numberOfPurchases || 0) + 1;
-    const timeUntilNextPurchase = calculateEstimate(
+    const daysUntilNextPurchase = calculateEstimate(
       listItem.timeFrame,
       latestInterval,
       numberOfPurchases,
     );
+    console.log('Time until next purchase: ', daysUntilNextPurchase);
 
     // FOR THE DEMO
     // ------------------------------------------------------------------------------------
@@ -73,7 +78,7 @@ const List = ({ token }) => {
       // start here - gives the number of days till next purchase
       const nextPurchaseDate = addMilliseconds(
         previouslyPurchasedDate,
-        timeUntilNextPurchase,
+        daysUntilNextPurchase,
       );
       // actual next purchase date base on oneWeekAgo value
       const durationOfTime = formatDistance(
@@ -81,12 +86,12 @@ const List = ({ token }) => {
         new Date(nextPurchaseDate),
       );
       // For PR review only
-      console.log('next purchase date: ', nextPurchaseDate);
-      console.log('duration of time: ', durationOfTime);
+      // console.log('next purchase date: ', nextPurchaseDate);
+      // console.log('duration of time: ', durationOfTime);
 
       queryCollection.update({
         lastPurchased: currentDate,
-        timeUntilNextPurchase: timeUntilNextPurchase,
+        daysUntilNextPurchase: daysUntilNextPurchase,
         numberOfPurchases: numberOfPurchases,
         durationOfTime: durationOfTime, // For PR review purposes only
         nextPurchaseDate: nextPurchaseDate, // For PR review purposes only
@@ -95,7 +100,7 @@ const List = ({ token }) => {
       // This else block is used for review purposes only
       queryCollection.update({
         lastPurchased: currentDate,
-        timeUntilNextPurchase: timeUntilNextPurchase,
+        daysUntilNextPurchase: daysUntilNextPurchase,
         numberOfPurchases: numberOfPurchases,
       });
     }
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -2,8 +2,8 @@ import React from 'react';
 import useFirestore from '../hooks/useFirestore';
 import { db } from '../lib/firebase';
 import Error from './Error';
-//IMPORT ESTIMATE.JS
 import calculateEstimate from './../lib/estimates';
+// import { formatDistance, formatDuration, addMilliseconds, toDate, getTime } from 'date-fns';
 
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
@@ -12,12 +12,12 @@ const List = ({ token }) => {
     if (purchasedDate === null) {
       return false;
     }
-    // changed to using Date.now(), time given in milliseconds
+    // time given in milliseconds
     const today = Date.now();
     // 24 hours in milliseconds
     const addOneDay = 86400000;
     // Add 24 hours to purchased
-    const oneDayAfterPurchase = purchasedDate + 1000; //CHANGE 1000 BACK TO ADDONEDAY
+    const oneDayAfterPurchase = purchasedDate + addOneDay;
 
     // If today > oneDayAfterPurchase at least 24 hours have passed, return false to uncheck box
     if (today >= oneDayAfterPurchase) {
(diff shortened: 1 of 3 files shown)


src/components/List.js
```diff
@@ -26,10 +28,44 @@ const List = ({ token }) => {
...
+    // const exampleNumberOfPurchases = 7;
+    // const demoEstimate = calculateEstimate(
+    //   targetDoc[0].timeFrame,
+    //   demoLatestInterval,
+    //   exampleNumberOfPurchases,
+    // );
+    // demo estimate tells how long from purchase date until we're expected to buy it again
+    // console.log('Demo estimate :', addMilliseconds(previouslyPurchasedDate, demoEstimate));
+    // ------------------------------------------------------------------------------------
+    // END OF DEMO
```
LINDGRENBA, line 63: This code is specifically for our demo on Sunday. We will install the date-fns dependency to demo the times in the console. However, this is not necessary for the functionality and will be removed before merging.

src/components/List.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import useFirestore from '../hooks/useFirestore';
 import { db } from '../lib/firebase';
 import Error from './Error';
+import calculateEstimate from './../lib/estimates';
+// import { formatDistance, formatDuration, addMilliseconds, toDate, getTime } from 'date-fns';
```
LINDGRENBA, line 6: These date-fns import statements are for the demo only and will be removed before merging into main.

LINDGRENBA (approved):

LINDGRENBA:
One question @veeteeran and I had is that we are reusing the code `targetDoc[0]` several times. Should we refactor to store this in a variable and use the variable in its place?

laurenmbeatty:
> One question @veeteeran and I had is that we are reusing the code `targetDoc[0]` several times. Should we refactor to store this in a variable and use the variable in its place?

I vote yes.

rachelrohrbach (approved):
Looks good, way to go on giving our app the "smart" shopping list functionality! I tried the Netlify demo and it seems to be working well and updating the firestore collection accordingly. I appreciated the comments in the code!  
And I vote yes as well to storing targetDoc[0] as a variable since you're reusing it several times in the code.

DeeDowns (approved):
I also vote for creating a variable for `targetDoc[0]`
Great job figuring this issue, testing it out and it worked perfectly. This issue seemed challenging but you guys did well 👌

src/components/List.js
```diff
@@ -55,6 +94,7 @@ const List = ({ token }) => {
                 id={doc.id}
                 onChange={handleCheckbox}
                 checked={checkPurchasedDate(doc.lastPurchased)}
+                disabled={checkPurchasedDate(doc.lastPurchased)}
```
veeteeran, line 97: @LINDGRENBA and I think `checked` and `disabled` follow the same logic so passing the `checkedPurchaseDate` function to `disabled` is the route we took. Once checked the checkbox is disabled so a user can't continually tap the checkbox and update the `lastPurchased` date

src/components/List.js
```diff
@@ -26,10 +28,47 @@ const List = ({ token }) => {
   };
 
   const handleCheckbox = async (event) => {
-    const currentDate = Date.now(); // changed to using Date.now(), time given in milliseconds
+    const currentDate = Date.now(); // time given in milliseconds
     const queryCollection = await db.collection(token).doc(event.target.id);
+    // filteredListItem is our current snapshot of the doc
+    const filteredListItem = docs.filter(
+      (doc) => doc.id === queryCollection.id,
+    );
```
laurenmbeatty, line 36: I'm wondering, could you just skip the call the db, and just use the `event.target.id` in place of `queryCollection.id`?  Little less costly, I think.

laurenmbeatty: and then below at the `queryCollection.update` part...you could just do:

`db.collection(token).doc(event.target.id).update.....`

src/components/List.js
```diff
@@ -26,10 +28,47 @@ const List = ({ token }) => {
...
+    // console.log('Demo estimate :', addMilliseconds(previouslyPurchasedDate, demoEstimate));
+    // ------------------------------------------------------------------------------------
+    // END OF DEMO
+
     queryCollection.update({
       lastPurchased: currentDate,
+      estimatedNextPurchase: estimate,
+      numberOfPurchases: numberOfPurchases,
     });
   };
```
laurenmbeatty, line 73: This is a super-tricky one, and I'm trying to understand it, but right now, the estimatedNextPurchase is being saved as milliseconds, correct? Right now this is one of the values that I saved: 1612403611806, which converts to around 18,000 days or 01/13/53065. 

image

I wonder if the three params for the estimates function above need to be 

1. `estimatedNextPurchase || timeFrame` in _days_ (use timeFrame if there hasn't been a calculated value yet
2. latest interval should be the distance between last purchase and this purchase in _days_
3. Number of purchases
Then it should return a number in _days_

I 100% could be wrong.

src/components/List.js
```diff
@@ -26,10 +28,47 @@ const List = ({ token }) => {
...
+    // const exampleNumberOfPurchases = 7;
+    // const demoEstimate = calculateEstimate(
+    //   targetDoc[0].timeFrame,
+    //   demoLatestInterval,
+    //   exampleNumberOfPurchases,
+    // );
+    // demo estimate tells how long from purchase date until we're expected to buy it again
+    // console.log('Demo estimate :', addMilliseconds(previouslyPurchasedDate, demoEstimate));
+    // ------------------------------------------------------------------------------------
+    // END OF DEMO
```
laurenmbeatty, line 66: Ooo..cool.

src/components/List.js
laurenmbeatty: Hmm...If I leave off the last three digits...then it's OK? You know what would be awesome...could you display the estimated value in the # of days next to the item name? Then it would be easier to understand and your reviewer would be super grateful : ). 👍

src/components/List.js
```diff
@@ -26,10 +28,47 @@ const List = ({ token }) => {
...
+    const filteredListItem = docs.filter(
+      (doc) => doc.id === queryCollection.id,
+    );
+    const listItem = filteredListItem[0];
+    const previouslyPurchasedDate = listItem.lastPurchased;
+    const latestInterval = currentDate - previouslyPurchasedDate;
+    // If item has not been purchased set to 1, else increment by 1
+    const numberOfPurchases = !listItem.numberOfPurchases
+      ? 1
+      : (listItem.numberOfPurchases += 1);
```
MikeBLambert, line 43: In functional programming, which is the paradigm that React more or less ascribes to, it is thought of as best practice to avoid mutating data as it can cause non-deterministic behavior really hard to catch bugs. 

In this case, we are mutating `listItem.numberOfPurchases` with `+=`. I think that we can just replace `+=` with `+` as we're really just looking to assign the value of `listItem.numberOfPurchases + 1` to `numberOfPurchases`

Here's a good little article about immutability in functional programming - https://dev.to/mpodlasin/functional-programming-in-js-part-ii-immutability-vanilla-js-immutable-js-and-immer-2ccm

src/components/List.js
MikeBLambert: Nice! Love the idea of having something like this ready to help out with the demo. 
One thought that I had is that you could create a demo branch where you would keep this code, then you could merge the intended code changes before the demo without merging this demo code and also have something socked away that you can use for demo purposes.

src/components/List.js
LINDGRENBA: That's a great idea, and maybe we could add a screenshot of some of the values we're checking with those comments, too, so you can see them before our demo.

src/components/List.js
LINDGRENBA: Yes! Of course 🤦 that's React's whole thing... I think it totally escaped us that that's what we were doing there. Thank you for pointing it out.

src/components/List.js
LINDGRENBA: We added functionality to display the duration of time until the next purchase, but had some issues trying to display the actual date of the next purchase. After spending a reasonable amount of time trying to get it to get it to work we decided to move on. 
timeTillNextPurchase-consoleLogs

src/components/List.js
LINDGRENBA: When we change `queryCollection.id` to `event.target.id`  or just try to console.log the value of `event.target.id`, we get the following error :
target-event-id-error

chrisdopuch:
Hey @veeteeran @LINDGRENBA I tried QA testing the app on the netlify demo, but I ran into an issue. After adding an item to my list, I am unable to check the checkbox. Popping open the terminal, I see the following error show up when I try to check the box:

src/components/List.js
```diff
@@ -26,12 +34,67 @@ const List = ({ token }) => {
...
   const handleCheckbox = async (event) => {
-    const currentDate = Date.now(); // changed to using Date.now(), time given in milliseconds
+    const currentDate = Date.now(); // time given in milliseconds
     const queryCollection = await db.collection(token).doc(event.target.id);
+    // listItem is our current snapshot of the doc
+    const listItem = docs.filter((doc) => doc.id === queryCollection.id)[0];
+    const previouslyPurchasedDate = listItem.lastPurchased;
+    const latestInterval = currentDate - previouslyPurchasedDate;
+    // If item has not been purchased set to 1, else increment by 1
+    const numberOfPurchases = !listItem.numberOfPurchases
```
chrisdopuch, line 44: Way to check for the `numberOfPurchases` being undefined 👍 

In many programming languages, there's a thing called a "guard" statement that allows you to express this kind of logic in a more compact way. JS has this feature through the use of our boolean operators (e.g. `||` or `&&`).

You can use those operators to check for certain values truthiness just like we're doing here, but with a shorter syntax.
That would look like:
```
    const numberOfPurchases = (listItem.numberOfPurchases || 0) + 1;
```

This is just a code style kind of thing, so definitely optional. I just wanted to make you aware of it, if you didn't already know! Here's a blog post that goes into more detail if you're curious how it works: https://seanmonstar.com/post/707078771/guard-and-default-operators

src/components/List.js
```diff
@@ -55,8 +118,10 @@ const List = ({ token }) => {
                 id={doc.id}
                 onChange={handleCheckbox}
                 checked={checkPurchasedDate(doc.lastPurchased)}
+                //disabled={checkPurchasedDate(doc.lastPurchased)} // Commented out for PR purposes only
               />
               {doc.itemName}
+              <p>Time until next purchase {doc.durationOfTime}</p>
```
chrisdopuch, line 124: It appears this value is missing when I QA tested on netlify:

chrisdopuch (commented):
Nice work so far y'all, getting close! 👏 
I'll come back later and approve once the demo is working, just ping me once you've fixed the issue @veeteeran @LINDGRENBA 😄

src/components/List.js
veeteeran: I did not know about this. Thanks!

src/components/List.js
veeteeran: @chrisdopuch The above issue has been fixed. Thanks for hopping on Zoom with me earlier. Things seem to be working 🤞🏽 😅

LINDGRENBA: Thanks for taking care of that! Also, @chrisdopuch , those are some great list items.

veeteeran: Everyone needs 🧦 🧦  and gummie bears!

LINDGRENBA:
@chrisdopuch it looks like the Netlify issue is resolved, at least when I try it on my end. Let us know if it's still not working for you!

src/components/List.js
```diff
@@ -121,7 +130,11 @@ const List = ({ token }) => {
                 //disabled={checkPurchasedDate(doc.lastPurchased)} // Commented out for PR purposes only
               />
               {doc.itemName}
-              <p>Time until next purchase {doc.durationOfTime}</p>
+              {doc.numberOfPurchases > 0 ? (
+                <p>Time until next purchase {doc.durationOfTime}</p>
+              ) : (
+                <p>You haven't purchased {doc.itemName} yet</p>
```
chrisdopuch, line 136: Nice, this is a great UX improvement! 👏

src/components/List.js
```diff
@@ -121,7 +130,11 @@ const List = ({ token }) => {
                 //disabled={checkPurchasedDate(doc.lastPurchased)} // Commented out for PR purposes only
               />
               {doc.itemName}
-              <p>Time until next purchase {doc.durationOfTime}</p>
+              {doc.numberOfPurchases > 0 ? (
+                <p>Time until next purchase {doc.durationOfTime}</p>
```
chrisdopuch, line 134: @veeteeran @LINDGRENBA the first time I check the checkbox, the "time until next purchase" text shows up but it doesn't display a value. Should it be displaying something?
purchase

Once I click it a second time, then it starts to show the value

chrisdopuch (approved):
This is looking good you two! 😄 🎉 
I had one small question about if the value should be shown for the first click of the checkbox
But otherwise, it's working great now!

veeteeran:
@chrisdopuch I realized the condition is written to check if numberOfPurchases is greater than 0. We can change it to greater than 1 and display a message saying "More info need to calculate next purchase"

or something less wordy? 😄

src/components/List.js
LINDGRENBA: Hi @chrisdopuch . One thing I wasn't sure was clear is that the value is only there to make it easier for reviewers to see that it's working and won't be part of the final UI/UX. Lauren suggested it in one of the previous comments. I still think this was a good thing to point out because it may be an indicator of the code not functioning as we expected, but it won't be part of the final UI.... unless the team likes it and wants to keep it in, but I think we deal with organizing the items by this value in one of next week's issues and there's a different approach 👍 😄

MikeBLambert: I may be totally misunderstanding things (as I often do... ;-P), but it looks like right now we're saving to the DB a human readable form of the time to next purchase at the exact time that the user checked the checkbox. That means that that value will never change, even as time marches on. I believe that we may only need to save the estimated next purchase date and the number of times that the item has been purchased in the database. 

We should be able to derive the time until the next purchase from the distance between now and the estimated next purchase date and we can just calculate that on the fly via a function in our component. My general rule for determining whether or not something should be saved to a DB (with every rule there are always exceptions...) is that if I can derive it from other pre-existing values, then there's no need to save it. 

However, as you mentioned, this is just for demo purposes, so it's not a biggie either way. :-)

src/components/List.js
laurenmbeatty: Hmmm....I guess when I did it when playing around on your branch I grabbed the value of `event.target.id` right at the beginning of the function:

`const id = event.target.id`

laurenmbeatty:
When I look in Firestore I see:
image

The AC says: "When a purchase is recorded, the _estimated number of days_ until the next purchase date should be calculated and recorded in the database". 

Which of the fields in the db refers to the estimated number of days? I'm expecting to see a whole number value like: 16, 30, 3, 14, --- a number similar to the `timeFrame` field, but maybe the plan is to use one of those numbers and convert it into a value that reflects number of days.

src/components/List.js
MikeBLambert: Correcting myself slightly as @laurenmbeatty rightly pointed out we are supposed to be recording the estimated days until next purchase, not the next purchase date. I would take that to mean the number of days, which, if added onto the last purchase date, will give us our next purchase date. Thus, if we want to display how much time a user has at any given moment until they should be purchasing again, we would not want to directly display the value in the DB, as that won't change, but rather we would display the time between now and that next purchase date.

laurenmbeatty:
I've created a sandbox of the problem here, and the kind of values I would expect to pass to the function, and the kind of values I would expect to be returned and saved in firestore:
https://repl.it/join/lptucqyb-laurenbeatty

If you click "Run" it will give you the logged values in the shell: 
image

veeteeran:
Hi @laurenmbeatty. Thanks for pointing us in the right direction. We completely overlooked inputting values as whole numbers and we can definitely rename our variable to be more concise. I've found a function in date-fns `differenceInDays` that should give us the whole number of days to pass as the second argument.
`const latestInterval = differenceInDays(currentDate, lastPurchased);`


> // example if it was the first purchase, so there really wasn't a latest interval
> // I'm passing in whatever the timeFrame was to latestInterval value instead, since getting the latestInterval between the current purchase date and null seems weird

I see what you mean here. we should be checking if `numberOfPurchases < 2` set the latestInterval to the timeFrame else set latestInterval to the whole number of days as figured above

> // What if you're purchasing it on the same day??? what does that give you?
For this question, your testing shows giving 0 as the second argument works. This is great since `differenceInDays` returns 0 if the difference is under 24 hours. I don't think this would be an issue if zero didn't work since we disable the checkbox. At the moment we have that line commented out for the purpose of reviewing.

Sorry, that was a long reply. I hope it wasn't too confusing. 😅  I'll push code as soon as I can

src/components/List.js
veeteeran: @MikeBLambert I agree with what @laurenmbeatty and you are saying about the values saved in the database. In the database we want to save `numberOfPurchases` and `daysUntilNextPurchase`.

What is time? 🤣

src/components/List.js
```diff
@@ -26,12 +28,52 @@ const List = ({ token }) => {
...
+    // listItem is our current snapshot of the doc
+    const listItem = docs.filter((doc) => doc.id === queryCollection.id)[0];
+    const previouslyPurchasedDate = listItem.lastPurchased;
+
+    /* latestInterval is the number of full days between the two dates
+     * or the timeFrame if the function returns NaN
+     */
+    const latestInterval =
+      differenceInDays(currentDate, previouslyPurchasedDate) ||
+      listItem.timeFrame;
```
laurenmbeatty, line 42: Nice!

src/components/List.js
```diff
@@ -26,12 +28,52 @@ const List = ({ token }) => {
...
+
+    // If item has not been purchased set to 1, else increment by 1
+    const numberOfPurchases = (listItem.numberOfPurchases || 0) + 1;
+    const daysUntilNextPurchase = calculateEstimate(
+      listItem.timeFrame,
```
laurenmbeatty, line 47: This is my only question. My thought for this calculation is that it only uses the `timeFrame` value on the very first purchase, and then subsequently uses the `daysUntilNextPurchase` value. I think the function expects "The last estimated purchase interval", so it makes sense to me to do that because the calculation gets more and more refined as you purchase that item more and more.  What do you all think?

laurenmbeatty (approved):
I just had one more question, but GOOD JOBBBBBB!!!!!!!

src/components/List.js
veeteeran: @laurenmbeatty 💯  We'll change it so the first purchase uses `timeFrame` and every one after uses `daysUntilNextPurchase` :

src/components/List.js
```diff
@@ -26,12 +28,66 @@ const List = ({ token }) => {
...
+   * Used below when rendering List
+   */
+  const getDaysUntilNextPurchase = (doc) => {
+    const nextPurchasedDate = addDays(doc.lastPurchased, doc.timeFrame);
+
+    /* Added 1 here since we're working with full days
+     * if user taps a checkbox and the timeFrame was 7 days, 6 would have been displayed
+     */
+    const daysUntilNextPurchase =
+      differenceInDays(nextPurchasedDate, Date.now()) + 1;
```
veeteeran, line 87: This is the documentation on the [differenceInDays](https://date-fns.org/v2.17.0/docs/differenceInDays) function from date-fns

MikeBLambert (approved):
Great job on a really tough issue! Looks great!

PR #32, opened 2021-02-07
Bl vt estimate next purchase

veeteeran:
Hi all. I made a mistake while working with the merge conflict. I think it was in the return statement starting on line 136. Missing parenthesis perhaps? The app is compiling and running on my machine. The filtering looks good and so does the estimate next purchase date. I've pushed to the BL-VT branch in case something else is funny. I haven't reverted the main branch. Lesson learned. Don't try to fix a merge conflict before going to bed. 😦  😴

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -92,9 +92,11 @@ const List = ({ token }) => {
     /* Added 1 here since we're working with full days
      * if user taps a checkbox and the timeFrame was 7 days, 6 would have been displayed
      */
-    const daysUntilNextPurchase =
-      differenceInDays(nextPurchasedDate, Date.now()) + 1;
-    return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
+    const daysUntilNextPurchase = differenceInDays(
+      nextPurchasedDate,
+      Date.now(),
+    );
+    return daysUntilNextPurchase > 0 ? daysUntilNextPurchase + 1 : 0;
   };
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/List.js
```diff
@@ -101,42 +92,22 @@ const List = ({ token }) => {
...
-                ?.toLowerCase()
-                ?.includes(searchInput.toLowerCase().trim()),
-            )
-            ?.map((doc) => (
+          docs.map((doc) => {
+            return (
               <li key={doc.id}>
                 <input
                   type="checkbox"
-                  aria-label="purchased-checkbox"
```
veeteeran, line 139: I noticed I don't have this line. Should we put it back in? Looks like it will now be on line 140.

rachelrohrbach: I added this for accessibility, so yes if we could put it back in.

src/components/List.js
```diff
@@ -132,29 +132,29 @@ const List = ({ token }) => {
...
-                  <p>You haven't purchased {doc.itemName} yet.</p>
-                )}
-              </li>
-            );
-          })}
+            ?.map((doc) => {
+              return (
+                <li key={doc.id}>
+                  <input
+                    type="checkbox"
```
laurenmbeatty, line 139:
```suggestion
                    type="checkbox"
                    aria-label="purchased-checkbox"
```

laurenmbeatty (approved):
Yep! Looks good 👍 ! You can just commit the aria-label line right here in gh.

PR #34, opened 2021-02-08
Bl vt estimate next purchase

LINDGRENBA:
## Related Issue
[#11](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/11) 

## Acceptance Criteria
- [ ] When a purchase is recorded, the estimated number of days until the next purchase date should be calculated and recorded in the database

## Type of Changes

|   ✓ | :hammer: Refactoring       |

### Before
snapshot of list

### After
list-UI-week5

* Otherwise from your terminal, pull down this branch with git pull origin `BL-VT-estimate-next-purchase` and check that branch out with `git checkout BL-VT-estimate-next-purchase`
* To install the date-fns dependency, enter `npm install --save date-fns` into your terminal
* Use `npm start` to launch the app.
* Create a new list and add items or if a list exists click the List button
* Tap a check box of any item on the list - confirm that the message for the corresponding list item updates with an estimated next purchse
* Open Firebase and confirm the fields `timeFrame`, `lastPurchased` and `numberOfPurchases` are updated

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -126,7 +126,7 @@ const List = ({ token }) => {
                     id={doc.id}
                     onChange={handleCheckbox}
                     checked={checkPurchasedDate(doc.lastPurchased)}
-                    //disabled={checkPurchasedDate(doc.lastPurchased)} // Commented out for PR purposes only
+                    disabled={checkPurchasedDate(doc.lastPurchased)}
                   />
                   {doc.itemName}
                   {doc.numberOfPurchases > 0 ? (
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -58,24 +58,6 @@ const List = ({ token }) => {
       numberOfPurchases,
     );
 
-    // FOR THE DEMO
-    // ------------------------------------------------------------------------------------
-    // update day of demo
-    // const oneWeekAgo = getTime(new Date(2021, 0, 26, 11, 30, 30));
-    // oneWeekAgo simulates the last date of purchase for the item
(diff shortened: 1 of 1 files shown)


src/components/List.js
```diff
@@ -92,11 +74,9 @@ const List = ({ token }) => {
...
      * if user taps a checkbox and the timeFrame was 7 days, 6 would have been displayed
      */
-    const daysUntilNextPurchase = differenceInDays(
-      nextPurchasedDate,
-      Date.now(),
-    );
-    return daysUntilNextPurchase > 0 ? daysUntilNextPurchase + 1 : 0;
+    const daysUntilNextPurchase =
+      differenceInDays(nextPurchasedDate, Date.now()) + 1;
+    return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
```
laurenmbeatty, line 79: Is this always greater than zero because you're adding 1 on line 78?

LINDGRENBA: I believe that this is what @veeteeran was talking about where if, for example, the value we get from `differenceInDays(nextPurchasedDate, Date.now( ))` should be 7 days, because it automatically starts counting down, the second we have even a millisecond less than 7 it rounds it all the way down to 6 days. So to counter that almost instantaneous action of rounding down, we add 1 to the number we get from `differenceInDays(nextPurchasedDate, Date.now( ))`

laurenmbeatty (approved):
LGTM 👍

PR #35, opened 2021-02-08
Add sorting of shopping list

### LINKED ISSUE #13: 12. As a user, I want to view a list of my shopping list items in order of how soon I am likely to need to buy each of them again so that it’s clear what I need to buy soon.
Possible item states are as follows:

* Need to buy soon (fewer than 7 days)
* Need to buy kind of soon (between 7 & 30 days)
* Need to buy not soon (more than 30 days)
* Inactive (when there’s only 1 purchase in the database or the purchase is really out of date [the time that has elapsed since the last purchase is 2x what was estimated])

AC:

* Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive
* Items should be sorted by the estimated number of days until next purchase
* Items with the same number of estimated days until next purchase should be sorted alphabetically
* Items in the different states should be described distinctly when read by a screen reader

rachelrohrbach:
## Description

This code sorts the user's list by the estimated number of days until next purchase and sorts those items with the same number of estimated days alphabetically and displays the sorted items as visually distinct with distinct descriptions for screen readers. 

## Related Issue
Closes #13 

## Acceptance Criteria

- [x] Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive
- [x] Items should be sorted by the estimated number of days until next purchase
- [x] Items with the same number of estimated days until next purchase should be sorted alphabetically
- [x] Items in the different states should be described distinctly when read by a screen reader

## Type of Changes

### Before

Screen Shot 2021-02-09 at 3 20 38 PM

### After

image

netlifyPreview

2. Otherwise from your terminal, pull down this branch with git pull origin RR-VT-sort-shopping list and check that branch out with git checkout origin RR-VT-sort-shopping list
   * The list should be sorted in descending order according to `daysUntilNextPurchase` and if two items have the same `daysUntilNextPurchase`  then by descending alphabetical order.

### DIFF
--- a/src/components/List.css
+++ b/src/components/List.css
@@ -1,13 +1,13 @@
 .soon {
-  background-color: green;
+  background-color: lightgreen;
 }
 
 .kind-of-soon {
-  background-color: orange;
+  background-color: peachpuff;
 }
 
 .not-too-soon {
-  background-color: blue;
+  background-color: lightblue;
 }
 
 .inactive {
--- a/src/components/List.css
+++ b/src/components/List.css
@@ -1,0 +1,15 @@
+.soon {
+  background-color: green;
+}
+
+.kind-of-soon {
+  background-color: orange;
+}
+
+.not-too-soon {
+  background-color: blue;
+}
+
+.inactive {
+  background-color: lightgrey;
+}
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -4,7 +4,7 @@ import { db } from '../lib/firebase';
 import Error from './Error';
 import calculateEstimate from './../lib/estimates';
 import { differenceInDays, addDays } from 'date-fns';
-import './List.css';
+// import './List.css';
 
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -4,13 +4,12 @@ import { db } from '../lib/firebase';
 import Error from './Error';
 import calculateEstimate from './../lib/estimates';
 import { differenceInDays, addDays } from 'date-fns';
-// import './List.css';
+import './List.css';
 
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
   const [searchInput, setSearchInput] = useState('');
 
-  // const sortedList = Array.from(docs);
   const sortedList = docs.sort((a, b) => {
     if (a.timeFrame < b.timeFrame) {
       return -1;
@@ -27,12 +26,6 @@ const List = ({ token }) => {
     }
     return 0;
   });
-  console.log(sortedList);
-  // const o = {id: doc.itemName}
-  //  console.log(doc.id);
-  //   console.log(doc.itemName);
-
-  // console.log(sortedList);
 
   const handleSearchChange = (event) => {
     setSearchInput(event.target.value);
@@ -104,18 +97,18 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
-  // const backgroundColor = (timeFrame) => {
-  //   console.log(timeFrame);
-  //   if (timeFrame < 7) {
-  //     return "soon";
-  //   }
-  //   if (timeFrame >= 7 && timeFrame < 30) {
-  //     return "kind-too-soon";
-  //   }
-  //   if (timeFrame >= 30) {
-  //     return "not-too-soon"
-  //   }
-  // }
+  const backgroundColor = (sortedList) => {
+    const daysUntilNextPurchase = getDaysUntilNextPurchase(sortedList);
+    if (!sortedList.numberOfPurchases || sortedList.numberOfPurchases < 2) {
+      return 'inactive';
+    } else if (daysUntilNextPurchase < 7) {
+      return 'soon';
+    } else if (daysUntilNextPurchase >= 7 && daysUntilNextPurchase < 30) {
+      return 'kind-of-soon';
+    } else if (daysUntilNextPurchase >= 30) {
+      return 'not-too-soon';
+    }
+  };
 
   return (
     <div>
@@ -154,10 +147,7 @@ const List = ({ token }) => {
             )
             ?.map((sortedList) => {
               return (
-                <li
-                  key={sortedList.id}
-                  // className={backgroundColor(sortedList.timeFrame)}
-                >
+                <li key={sortedList.id} className={backgroundColor(sortedList)}>
                   <input
                     type="checkbox"
                     aria-label="purchased-checkbox"
@@ -185,17 +175,3 @@ const List = ({ token }) => {
 };
 
 export default List;
-
-// Issue 12:
-// Possible item states are as follows:
-
-// Need to buy soon (fewer than 7 days)
-// Need to buy kind of soon (between 7 & 30 days)
-// Need to buy not soon (more than 30 days)
-// Inactive (when there’s only 1 purchase in the database or the purchase is really out of date [the time that has elapsed since the last purchase is 2x what was estimated])
-
-// Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive
-// Items should be sorted by the estimated number of days until next purchase
-// AC:
-// Items with the same number of estimated days until next purchase should be sorted alphabetically
-// Items in the different states should be described distinctly when read by a screen reader
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -4,10 +4,35 @@ import { db } from '../lib/firebase';
 import Error from './Error';
 import calculateEstimate from './../lib/estimates';
 import { differenceInDays, addDays } from 'date-fns';
+import './List.css';
 
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
   const [searchInput, setSearchInput] = useState('');
+
+  // const sortedList = Array.from(docs);
+  const sortedList = docs.sort((a, b) => {
+    if (a.timeFrame < b.timeFrame) {
+      return -1;
+    }
+    if (a.timeFrame > b.timeFrame) {
+      return 1;
+    }
+    if (a.timeFrame === b.timeFrame) {
+      if (a.itemName < b.itemName) {
+        return -1;
+      } else {
+        return 1;
+      }
+    }
+    return 0;
+  });
+  console.log(sortedList);
+  // const o = {id: doc.itemName}
+  //  console.log(doc.id);
+  //   console.log(doc.itemName);
+
+  // console.log(sortedList);
 
   const handleSearchChange = (event) => {
     setSearchInput(event.target.value);
@@ -79,6 +104,19 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  // const backgroundColor = (timeFrame) => {
+  //   console.log(timeFrame);
+  //   if (timeFrame < 7) {
+  //     return "soon";
+  //   }
+  //   if (timeFrame >= 7 && timeFrame < 30) {
+  //     return "kind-too-soon";
+  //   }
+  //   if (timeFrame >= 30) {
+  //     return "not-too-soon"
+  //   }
+  // }
+
   return (
     <div>
       <h1>List</h1>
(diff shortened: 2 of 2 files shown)


src/components/List.js
```diff
@@ -114,26 +151,31 @@ const List = ({ token }) => {
...
                   />
-                  {doc.itemName}
-                  {doc.numberOfPurchases > 0 ? (
+                  {sortedList.itemName}
+                  {sortedList.numberOfPurchases > 0 ? (
                     <p>
-                      Time until next purchase: {getDaysUntilNextPurchase(doc)}{' '}
-                      days
+                      Time until next purchase:{' '}
+                      {getDaysUntilNextPurchase(sortedList)} days. Purchased{' '}
```
rachelrohrbach, line 174: We added this UI in for now to help us check that it's sorting the list as we intended.

src/components/List.js
```diff
@@ -114,26 +151,31 @@ const List = ({ token }) => {
...
                 ?.includes(searchInput.toLowerCase().trim()),
             )
-            ?.map((doc) => {
+            ?.map((sortedList) => {
               return (
-                <li key={doc.id}>
+                <li
+                  key={sortedList.id}
+                  className={backgroundColor(sortedList)}
+                  aria-label={backgroundColor(sortedList)}
```
rachelrohrbach, line 159: We're able to dynamically assign the `backgroundColor` and `aria-label` with the backgroundColor function. 🌈

src/components/List.js
```diff
@@ -79,6 +80,42 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  const sortedList = docs.sort((a, b) => {
```
veeteeran, line 83: Here's a good resource we found on the [sort function](https://www.javascripttutorial.net/javascript-array-sort/)

src/components/List.js
veeteeran: 🦄

src/components/List.js
LINDGRENBA: I really like that you linked to the documentation here! It made it really easy for me to quickly review how the sort method works! Thank you.

src/components/List.js
```diff
@@ -79,6 +80,39 @@ const List = ({ token }) => {
...
+    if (aDaysUntilNextPurchase === bDaysUntilNextPurchase) {
+      if (a.itemName < b.itemName) {
+        return -1;
+      } else {
+        return 1;
+      }
+    }
+    return 0;
+  });
+
```
LINDGRENBA, line 102: Wow! I am super impressed. The code for the `sortedList` and `backgroundColor` code blocks is slick! There's thing I'm wondering about - I might be misunderstanding the AC, but I think there may be an issue with whether or not the `sortedList` is getting re-sorted as the `daysUntilNextPurchase` value changes.

* Initially we use the radio button value of 7, 14 or 30 for our `timeFrame`.
* Then once the user begins to purchase the item, the `timeFrame` is based on `daysUntilNextPurchase` instead of the 7, 14 and 30 value <i>and</i> our estimate.js calculation gets "smarter".
* Question :  do our background colors change accordingly as the app gets smarter?

Here is a snapshot of the list `stroll ames knack` which might be part of my confusion : 
background-colors

If the background colors should be similar `according to how soon the item is expected to be bought again` <i>and</i> `items should be sorted by the estimated number of days until next purchase`, should items with similar backgrounds be grouped together and have similar estimated `time until next purchase`?

Is this possibly a situation where we would need something like a `useEffect` hook or some other form of state to reset the value of `sortedList` any time the docs are updated?

src/components/List.js
```diff
@@ -79,6 +80,39 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  const sortedList = docs.sort((a, b) => {
```
LINDGRENBA, line 83: Even though I see `sortedList` referenced, there is a message in my terminal that `sortedList is assigned a value, but never used`

sortedListVariableNotUsed

LINDGRENBA (approved):
* Items in the list are shown as visually distinct CHECK ✓ Nice job! 😄 

* Items with the same number of estimated days until next purchase should be sorted alphabetically
CHECK ✓ Nice job! 😄 

* Items in the different states should be described distinctly when read by a screen reader
screen reader on Windows 10 reads out ‘inactive’, ‘soon’, ‘kind of soon’ and ‘not too soon’
CHECK ✓ Nice job! 😄 

I think the only thing is whether or not the background colors need to change as the estimated time until next purchase changes?

src/components/List.js
rachelrohrbach: Yes, we noticed this same warning and were a little confused by it.

src/components/List.js
veeteeran: > Question : Are our background colors supposed to change accordingly as the app gets smarter?

Good question. We're using the `getDaysUntilNextPurchase` function to inside of the `backgroundColor` function. We think the colors should update properly since `getDaysUntilNextPurchase` uses `timeFrame` which now gets the return from `calculateEstimate`

> should items with similar backgrounds be grouped together and have similar estimated time until next purchase? I get the impression that this might be the case due to Stacie's demo around second 0:25

We were unsure of how to handle the sorting of `inactive` items on the list. I see what you're saying about the demo and how all inactive items are at the end of the list. At the moment active and inactive items are grouped together by time frame. This is definitely a situation where we are looking for input from the team. 😄 

> Is this possibly a situation where we would need something like a useEffect hook or some other form of state to reset the value of sortedList any time the docs are updated?

We considered using useEffect but wasn't sure if it was necessary.  Also another place where we are looking for input from the team. 😄

DeeDowns: Currently testing it out with netlify preview, the sorting is really cool, it definitely has a smart feel to it as you check an item and the list is reordered. 
I get easily confused so it took me a while to wrap my head around the difference in the sorting and the background color lol. The items are sorted based on the number of days until the next purchase, and the background color is just to represent how soon a user initially expected to need the item again, right?

DeeDowns (approved):

src/components/List.js
```diff
@@ -79,6 +80,53 @@ const List = ({ token }) => {
...
+      if (a.itemName < b.itemName) {
+        return -1;
+      } else {
+        return 1;
+      }
+    }
+    return 0;
+  });
+
+  function isInactive(item) {
```
rachelrohrbach, line 110: By adding this helper function `isInactive` we can use the 'inactive' property as a boolean rather than just a string which we are able to use in the sorting function conditional logic to sort the inactive items to the end of the list.

src/components/List.js
```diff
@@ -107,33 +155,37 @@ const List = ({ token }) => {
       {errorMessage && <Error errorMessage={errorMessage} />}
 
       <ul style={{ listStyleType: 'none' }}>
-        {docs &&
-          docs
-            ?.filter((doc) =>
-              doc?.itemName
+        {sortedList &&
```
rachelrohrbach, line 158: Now we are using the `const sortedList` that we assigned back in our sort function which removes the console warning we were seeing previously.

src/components/List.js
```diff
@@ -107,33 +155,37 @@ const List = ({ token }) => {
...
+          sortedList
+            ?.filter((item) =>
```
rachelrohrbach, line 160: I think renaming doc to `item` is a bit more clear since here we are dealing with the individual list item(s).

src/components/List.js
rachelrohrbach: The latest code I pushed up now includes sorting of the inactive items.

src/components/List.js
rachelrohrbach, line 160: I think renaming `doc` to item is a bit more clear since here we are dealing with the individual list item(s).

src/components/List.css
```diff
@@ -0,0 +1,15 @@
+.soon {
+  background-color: green;
+}
+
+.kind-of-soon {
+  background-color: orange;
+}
+
+.not-too-soon {
+  background-color: blue;
```
chrisdopuch, line 10: Color contrast is an important aspect of accessibility on the web. https://www.a11yproject.com/posts/2015-01-05-what-is-color-contrast/

Could we adjust these color values to make the text more readable? Or adjust the text color (i.e. make it white instead of black)?

As it stands these are a little hard to read for me.
Imagine how much harder it would be for vision-impaired folks!

chrisdopuch (approved):
This is looking good, nice work y'all! Code is really solid, my only feedback was about the usability of the color contrasts.

src/components/List.css
laurenmbeatty: This was my main feedback as well. You could use an `rgba` value for the background color...the `a` allows you to add opacity of the color between 0 and 1.0.

laurenmbeatty: image
 You can use the dev tools color picker to figure out what that value could be.

src/components/List.js
```diff
@@ -107,33 +155,37 @@ const List = ({ token }) => {
...
                 ?.includes(searchInput.toLowerCase().trim()),
             )
-            ?.map((doc) => {
+            ?.map((item) => {
               return (
-                <li key={doc.id}>
+                <li
+                  key={item.id}
+                  className={backgroundColor(item)}
+                  aria-label={backgroundColor(item)}
```
laurenmbeatty, line 170: I think this needs a little more description name like : itemname ready to purchase kind of soon. Or something like that. Something that gives screen reader users a clear message. I know you have that text on the screen now for testing purposes, but eventually I'm guessing you won't.

laurenmbeatty (approved):
Hey, all. Looks good to me and meets the AC. For both of the functions you wrote with the conditionals, I would consider a `switch` statement. It's totally a personal preference though. When I find myself writing a bunch of `if`s, I reach for the switch. 🧹   

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

veeteeran:
> Hey, all. Looks good to me and meets the AC. For both of the functions you wrote with the conditionals, I would consider a `switch` statement. It's totally a personal preference though. When I find myself writing a bunch of `if`s, I reach for the switch. 🧹

Hi @laurenmbeatty . Rachel and I attempted using a switch to change the backGround color and came up with this.

Screen Shot 2021-02-11 at 6 39 10 PM

It's a little confusing for us. Is there something we're missing? Is there a way to give the case a range from 0-6?

Thanks!

veeteeran:
> Hey, all. Looks good to me and meets the AC. For both of the functions you wrote with the conditionals, I would consider a `switch` statement. It's totally a personal preference though. When I find myself writing a bunch of `if`s, I reach for the switch. 🧹
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

Hi  @laurenmbeatty Rachel and I are trying to use the switch statement and came up with this.

Screen Shot 2021-02-11 at 6 39 10 PM

We tried to write it this way but it's not working. 
Screen Shot 2021-02-11 at 6 49 16 PM


Is there something we're missing? Thanks!

laurenmbeatty:
> > Hey, all. Looks good to me and meets the AC. For both of the functions you wrote with the conditionals, I would consider a `switch` statement. It's totally a personal preference though. When I find myself writing a bunch of `if`s, I reach for the switch. 🧹
> > https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch
> 
> Hi @laurenmbeatty Rachel and I are trying to use the switch statement and came up with this.
> 
> Screen Shot 2021-02-11 at 6 39 10 PM
> 
> We tried to write it this way but it's not working.
> Screen Shot 2021-02-11 at 6 49 16 PM
> 
> Is there something we're missing? Thanks!

let me try and see if I can get something working. 👍

src/components/List.css
```diff
@@ -0,0 +1,15 @@
+.soon {
+  background-color: lightgreen;
+}
+
+.kind-of-soon {
+  background-color: peachpuff;
```
laurenmbeatty, line 6: yessssss!!!!!

MikeBLambert: wait, what??? that's a real color?

src/components/List.js
```diff
@@ -79,6 +80,53 @@ const List = ({ token }) => {
...
+      return 1;
+    }
+
+    if (aDaysUntilNextPurchase < bDaysUntilNextPurchase) {
+      return -1;
+    }
+    if (aDaysUntilNextPurchase > bDaysUntilNextPurchase) {
+      return 1;
+    }
+    if (aDaysUntilNextPurchase === bDaysUntilNextPurchase) {
```
MikeBLambert, line 100: I think that `if (aDaysUntilNextPurchase === bDaysUntilNextPurchase) {` may be unnecessary. If A is not less than B and A is not greater than B, then A must equal B, right?

MikeBLambert (approved):
Really great work. `.sort` always makes my brain hurt. Really great job working out that algorithm!
Oh also, what's up with all the coconut? Are we making piña coladas???

PR #36, opened 2021-02-09
Add delete functionality with error handling

LINDGRENBA:
## Description
This PR will allow the user to delete an item. When the user selects `delete` they will be shown a message to confirm that they actually want to delete the item from their list. This allows the user to delete items they no longer want to purchase and helps to declutter the list view.

## Related Issue
[#12](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/12)

## Acceptance Criteria
- [ ] User is able to delete an item from the shopping list
- [ ] Before deleting, prompt the user to confirm that they really want to delete the item to prevent accidental deletions
- [ ] Deletion should cause the associated record(s) in the database to be deleted

## Type of Changes

### Before
list-UI-week5

### After
ezgif com-video-to-gif (3)

deleteErrorMessage

* Otherwise from your terminal, pull down this branch with `git pull origin BL-DD-delete-item` and check that branch out with `git checkout BL-DD-delete-item`
* If you would like to join an existing list with lots of items to delete, clear any tokens in your `localStorage` and join the existing list with the token `stroll ames knack`
* Click the `delete` button next to one of the items
* Open Firebase and confirm the item is deleted from the collection
* If you would like to simulate an error, in `useFirestore.js` comment out `db.collection(token).doc(item.id)delete()` on line 30 and add `throw new Error( *backticks* Problem deleting ${item.itemName} *backticks* )` to line 31. When you are done testing, return these to their original states.

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -8,7 +8,6 @@ import { differenceInDays, addDays } from 'date-fns';
 const List = ({ token }) => {
   const { docs, errorMessage } = useFirestore(token);
   const [searchInput, setSearchInput] = useState('');
-  const [deleteError, setDeleteError] = useState('');
 
   const handleSearchChange = (event) => {
     setSearchInput(event.target.value);
@@ -83,14 +82,16 @@ const List = ({ token }) => {
   // Confirm that the user would like to delete this item
   const confirmDelete = (item) => {
     const confirmed = window.confirm(
-      'Are you sure you want to delete this item?',
+      `Are you sure you want to delete ${item.itemName}?`,
     );
     if (confirmed) {
       try {
         db.collection(token).doc(item.id).delete();
+        // testing purposes only
+        // throw new Error (`Problem deleting ${item.itemName}`)
       } catch (error) {
-        setDeleteError(`Problem deleting ${item.itemName}`);
-        console.log(error);
+        console.error(error);
+        alert(`Problem deleting ${item.itemName}`);
       }
     }
   };
@@ -121,7 +122,6 @@ const List = ({ token }) => {
       )}
 
       {errorMessage && <Error errorMessage={errorMessage} />}
-      {deleteError && <Error errorMessage={deleteError} />}
 
       <ul style={{ listStyleType: 'none' }}>
         {docs &&
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -144,3 +144,7 @@ const List = ({ token }) => {
 };
 
 export default List;
+
+// User is able to delete an item from the shopping list
+// Before deleting, prompt the user to confirm that they really want to delete the item to prevent accidental deletions
+// Deletion should cause the associated record(s) in the database to be deleted
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -79,6 +79,18 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  // Confirm that the user would like to delete this item
+  const confirmDelete = () => {
+    const confirmed = window.confirm(
+      'Are you sure you want to delete this item?',
+    );
+    if (confirmed) {
+      console.log('delete it');
+    } else {
+      console.log("don't delete it");
+    }
+  };
+
   return (
     <div>
       <h1>List</h1>
@@ -127,6 +139,7 @@ const List = ({ token }) => {
                     disabled={checkPurchasedDate(doc.lastPurchased)}
                   />
                   {doc.itemName}
+                  <button onClick={() => confirmDelete()}>Delete</button>
                   {doc.numberOfPurchases > 0 ? (
                     <p>
                       Time until next purchase: {getDaysUntilNextPurchase(doc)}{' '}
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -80,14 +80,13 @@ const List = ({ token }) => {
   };
 
   // Confirm that the user would like to delete this item
-  const confirmDelete = () => {
+  const confirmDelete = async (item) => {
+    const res = await db.collection(token).doc(item.id);
     const confirmed = window.confirm(
       'Are you sure you want to delete this item?',
     );
     if (confirmed) {
-      console.log('delete it');
-    } else {
-      console.log("don't delete it");
+      res.delete();
     }
   };
 
@@ -139,7 +138,7 @@ const List = ({ token }) => {
                     disabled={checkPurchasedDate(doc.lastPurchased)}
                   />
                   {doc.itemName}
-                  <button onClick={() => confirmDelete()}>Delete</button>
+                  <button onClick={() => confirmDelete(doc)}>Delete</button>
                   {doc.numberOfPurchases > 0 ? (
                     <p>
                       Time until next purchase: {getDaysUntilNextPurchase(doc)}{' '}
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -80,16 +80,17 @@ const List = ({ token }) => {
   };
 
   // Confirm that the user would like to delete this item
-  const confirmDelete = async (item) => {
-    const res = await db.collection(token).doc(item.id);
+  const confirmDelete = (item) => {
     const confirmed = window.confirm(
       'Are you sure you want to delete this item?',
     );
     if (confirmed) {
-      const didItemDelete = res.delete();
-      didItemDelete
-        ? console.log('Item was deleted')
-        : console.log('There was a problem');
+      try {
+        db.collection(token).doc(item.id).delete();
+      } catch (error) {
+        // <Error errorMessage={error}
+        console.log(error);
+      }
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/components/List.js
```diff
@@ -79,6 +80,21 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  // Confirm that the user would like to delete this item
+  const confirmDelete = (item) => {
+    const confirmed = window.confirm(
```
DeeDowns, line 85: We decided to use a confirm() pop up to prompt the user on whether or not they really want to delete the item. This can be changed to a styled modal or other pop up, but we didn't want to add too much styling prematurely without consulting the rest of the group.

src/components/List.js
```diff
@@ -79,6 +80,21 @@ const List = ({ token }) => {
...
+      'Are you sure you want to delete this item?',
+    );
+    if (confirmed) {
+      try {
+        db.collection(token).doc(item.id).delete();
+      } catch (error) {
+        setDeleteError(`Problem deleting ${item.itemName}`);
```
DeeDowns, line 92: We currently dont have a way of testing getting into the `catch` block. If anyone has any suggestions of how to do that (so the error message shows up when there is an actual error deleting the item) we would appreciate that.

src/components/List.js
```diff
@@ -127,6 +144,7 @@ const List = ({ token }) => {
                     disabled={checkPurchasedDate(doc.lastPurchased)}
                   />
                   {doc.itemName}
+                  <button onClick={() => confirmDelete(doc)}>Delete</button>
```
LINDGRENBA, line 147: While the project wireframe shows `delete` as a link, the resources we found advised the use of a button. 
[When to Use Buttons and Links](https://balsamiq.com/learn/articles/buttons-links/)
```
Links should never be used to change the state of an application. This means that clicking on a link should not add, change, or delete any data on the screen, So, a delete action should never be a link, nor should a save action.
```

src/components/List.js
laurenmbeatty: I wonder if you put a fake token in the `try` block if it would error:

`db.colleciont('fake').doc(item.id).delete();`

src/components/List.js
laurenmbeatty: agreed

veeteeran (approved):
Way to tackle this task. I was able to use the new feature in my end. Thanks for putting in that  note to get the function to error. Can you tell me why there's an error when deleting an item that hasn't been purchased? Nice work team!

rachelrohrbach (approved):
Great job, you two! The code is clean and works well.  I like that you used the confirm() pop up prompt for now and went for a delete button versus a link (and linked to documentation).

src/components/List.js
LINDGRENBA: That's a great idea!

LINDGRENBA:
@veeteeran, is that happening when you follow our testing directions?

veeteeran:
@LINDGRENBA It's when I make it error according to your screenshot.

DeeDowns:
> Way to tackle this task. I was able to use the new feature in my end. Thanks for putting in that note to get the function to error. Can you tell me why there's an error when deleting an item that hasn't been purchased? Nice work team!

@veeteeran I'm not getting that error when I delete an unpurchased item 🤔

src/components/List.js
laurenmbeatty, line 92: Might be a good idea to clear this error somewhere, if, for example, there ends up being an error and you make it to the catch block.

src/components/List.js
laurenmbeatty: Nice.

src/components/List.js
```diff
@@ -79,6 +80,21 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  // Confirm that the user would like to delete this item
+  const confirmDelete = (item) => {
+    const confirmed = window.confirm(
+      'Are you sure you want to delete this item?',
```
laurenmbeatty, line 86: As long as you're passing in the item, it would be easy to add that to your message to make it more specific.
```suggestion
      `Are you sure you want to delete ${item.itemName}?`,
```

laurenmbeatty (approved):
Love it! Great work. 💪 🎉 💯

src/components/List.js
laurenmbeatty: Maybe to keep the same strategy for now you could just alert the error, then you wouldn't need to clear it from the page, really.

LINDGRENBA: The second options sounds like a good short term fix and then if we're able we could tackle doing it the way we have it now, but then clearing it or resetting it to an empty string... @DeeDowns , what do you think?

LINDGRENBA:
> @LINDGRENBA It's when I make it error according to your screenshot.

Hey @veeteeran , did Dee's response answer your question? 😄

veeteeran:
Oh I see I thought the example  was only causing an error if the item had not been purchased. 😄

src/components/List.js
```diff
@@ -79,6 +80,21 @@ const List = ({ token }) => {
...
+    );
+    if (confirmed) {
+      try {
+        db.collection(token).doc(item.id).delete();
+      } catch (error) {
+        setDeleteError(`Problem deleting ${item.itemName}`);
+        console.log(error);
```
chrisdopuch, line 93: If you want to log out an error, I'd recommend using `console.error` so that it clearly shows up as an error in the console.

src/components/List.js
```diff
@@ -105,6 +121,7 @@ const List = ({ token }) => {
       )}
 
       {errorMessage && <Error errorMessage={errorMessage} />}
+      {deleteError && <Error errorMessage={deleteError} />}
```
chrisdopuch, line 124: Will there ever be multiple different errors at once? It might be more convenient to just reuse the existing error message functionality instead of repeating yourself for each kind of error type.

src/components/List.js
chrisdopuch: Great choice!

chrisdopuch (approved):
Great work y'all! I QA tested on the demo site and it worked nicely. I just had some comments about code reuse and console.error

src/components/List.js
```diff
@@ -79,6 +80,21 @@ const List = ({ token }) => {
     return daysUntilNextPurchase > 0 ? daysUntilNextPurchase : 0;
   };
 
+  // Confirm that the user would like to delete this item
+  const confirmDelete = (item) => {
+    const confirmed = window.confirm(
+      'Are you sure you want to delete this item?',
+    );
+    if (confirmed) {
+      try {
```
MikeBLambert, line 89: Way to error handle!!!

src/components/List.js
LINDGRENBA: Good question. I think right now the `errorMessage` on line 123 is coming from the `useFirestore` hook, so we weren't sure how to integrate any errors coming from our delete action to make the code re-usable. I feel like I'm missing something about how we could be using `useFirestore`...

MikeBLambert: One thing you could do is have useFirestore actually do full CRUD operations. You could add a `deleteDoc` function that looks something like 
```javascript
const deleteDoc = item =>
 try {
        db.collection(token).doc(item.id).delete();
      } catch (error) {
        setErrorMessage(`Problem deleting ${item.itemName}`);
     }
```
and return that from useFirestore. Then, inside of `List`, you can call `deleteDoc` inside of `confirmDelete`. You will need some way to clear the error as was mentioned elsewhere. You could return `setError` from `useFireStore` and use that to clear the error. Or you could do something fun like have the error just clear itself in a setTimeout so that it looks something like 
```javascript
  const deleteDoc = (doc) => {
    try {
      db.collection(collection).doc(doc.id).delete();
    } catch {
      setErrorMessage(`Problem deleting ${item.itemName}`);
      setTimeout(() => {
        setErrorMessage('');
      }, 3000);
    }
  };
```
Or you could play around with something like https://www.npmjs.com/package/react-alert. All sorts of fun options!

src/components/List.js
MikeBLambert: Alternatively, if you ever want to trigger a `catch` block just to smoke test your code, you can always just put `throw new Error("Some random error message")` in the `try` block and that will automatically send the code straight to the `catch` block.

MikeBLambert (approved):
It's working great! I left some suggestions for small improvements, but overall this looks wonderful.

src/components/List.js
DeeDowns: I played around with this and was able to get everything working, moving the delete functionality into the `useFirestore` hook and just calling the `deleteDoc` function inside the `confirmDelete` function. That way we are able to reuse the Error component, without setting new state in the case of a delete error.
Brittany and I discussed how the error message should be cleared, would it be better for the user to clear it themselves or for us to clear it with something like the setTimeout function you mentioned?

src/hooks/useFirestore.js
```diff
@@ -25,7 +25,21 @@ const useFirestore = (collection) => {
     };
   }, [collection]);
 
-  return { docs, errorMessage };
+  const deleteDoc = (item) => {
+    try {
+      db.collection(collection).doc(item.id).delete();
+      // testing purposes only
```
MikeBLambert, line 31: We may want to take these comments out.

PR #37, opened 2021-02-13
  Use switch for sortedList, refactor backgroundColor, add more descriptive aria-label

veeteeran:
## Description

This code refactors `sortedList` using a switch statement, conditions in `backgroundColor` are refactored, and the `aria-label` for the list items are more descriptive

## Acceptance Criteria

- [x] Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive
- [x] Items should be sorted by the estimated number of days until next purchase
- [x] Items with the same number of estimated days until next purchase should be sorted alphabetically
- [x] Items in the different states should be described distinctly when read by a screen reader

## Type of Changes

|   | :sparkles: New feature     |

### Before

Screen Shot 2021-02-12 at 6 22 21 PM

### After

Screen Shot 2021-02-12 at 6 23 13 PM

netlifyPreview

2. Otherwise from your terminal, pull down this branch with git pull origin RR-VT-sort-shopping list and check that branch out with git checkout origin RR-VT-sort-shopping list
   * The list should be sorted in descending order according to `daysUntilNextPurchase` and if two items have the same `daysUntilNextPurchase`  then by descending alphabetical order.

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -84,27 +84,19 @@ const List = ({ token }) => {
     const aDaysUntilNextPurchase = getDaysUntilNextPurchase(a);
     const bDaysUntilNextPurchase = getDaysUntilNextPurchase(b);
 
-    if (!isInactive(a) && isInactive(b)) {
-      return -1;
-    }
-    if (isInactive(a) && !isInactive(b)) {
-      return 1;
-    }
-
-    if (aDaysUntilNextPurchase < bDaysUntilNextPurchase) {
-      return -1;
-    }
-    if (aDaysUntilNextPurchase > bDaysUntilNextPurchase) {
-      return 1;
-    }
-    if (aDaysUntilNextPurchase === bDaysUntilNextPurchase) {
-      if (a.itemName < b.itemName) {
+    switch (true) {
+      case !isInactive(a) && isInactive(b):
         return -1;
-      } else {
+      case isInactive(a) && !isInactive(b):
         return 1;
-      }
+
+      case aDaysUntilNextPurchase < bDaysUntilNextPurchase:
+        return -1;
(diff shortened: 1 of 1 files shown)


MikeBLambert (approved):
Looks good!

src/components/List.js
```diff
@@ -116,15 +108,11 @@ const List = ({ token }) => {
...
     const daysUntilNextPurchase = getDaysUntilNextPurchase(item);
-    if (isInactive(item)) {
-      return 'inactive';
-    } else if (daysUntilNextPurchase < 7) {
-      return 'soon';
-    } else if (daysUntilNextPurchase >= 7 && daysUntilNextPurchase < 30) {
+    if (isInactive(item)) return 'inactive';
+    if (daysUntilNextPurchase < 7) return 'soon';
+    if (daysUntilNextPurchase >= 7 && daysUntilNextPurchase < 30)
       return 'kind-of-soon';
```
chrisdopuch, line 114: It might be nice to inline this return statement too if all the others are like that, but maybe this is prettier doing it for us, in which case I'd defer to the tool

src/components/List.js
```diff
@@ -167,7 +155,9 @@ const List = ({ token }) => {
                 <li
                   key={item.id}
                   className={backgroundColor(item)}
-                  aria-label={backgroundColor(item)}
+                  aria-label={`${
```
chrisdopuch, line 158: Nice!

chrisdopuch (approved):
Thanks for adding these improvements, looks good to me 👍

src/components/List.js
veeteeran: @chrisdopuch That's prettier doing it's thing 😄

PR #38, opened 2021-02-16
Bl delete token words

LINDGRENBA:
## Trigger Warning 
You do not need to view the removed words to approve the PR request. If you decide to navigate to the `files changed` view, you will see the triggering or offensive words that have been removed from the project. If you do not want to see these words, please avoid navigating to the `files changed` view.

## Description
This PR will remove triggering and offensive words from the `token.js` words array. The removal of these words is in alignment with TCL's mission of inclusion and the creation of a safe space for all.

## Related Issue
n/a

## Acceptance Criteria
n/a

## Type of Changes

|   ✓ | Other              |

### Before
No changes to the UI

### After
No changes to the UI

**Please only follow these steps if you are comfortable being exposed to triggering or offensive words.** To confirm that these words have been removed from the file you can 
1. pull the changes down locally using `git pull BL-delete-token-words` 
2. navigate to that branch with `git checkout BL-delete-token-words` 
3. use `ctrl+F` (on Windows) or `cmd+F` (on Mac) to pull up the search feature and enter any of the triggering words (which can be found in the `files changed` view in github.


.prettierignore
```diff
@@ -0,0 +1,5 @@
+# Ignore artifacts
+build
+
+# Ignore token file
+token.js
```
laurenmbeatty, line 5:
```suggestion
tokens.js
```
If this file doesn't appear to do anything, let's just not use it, I think. We can probably just delete it.

LINDGRENBA: Are you talking about the token.js file?

laurenmbeatty (approved):
maybe delete prettierignore, or try changing the name to `tokens.js`.

.prettierignore
LINDGRENBA: Hmm... I tried the .prettierignore, but maybe I didn't structure it properly. I'll give it another go and see if that works! Thank you for sharing that👍

LINDGRENBA: > If this file doesn't appear to do anything, let's just not use it, I think. We can probably just delete it.

Is this in reference to the `token.js` file?

laurenmbeatty: Sorry...i mean the prettier ignore. The prettierignore might also need to be in place before you make changes? Now that the file is already saved in a certain format, changing the prettierignore won't revert it back to it's previous formatting.

laurenmbeatty: For this one I think we can just delete `prettierignore` file, and then leave `tokens.js` as it currently stands in this PR. Not a huge deal.

LINDGRENBA: Oh! The `.prettierignore` file! Sorry, I think I got confused there. I thought you were giving a suggestion on how to make it work. I see now that you are referencing what is already there, but not functioning and suggesting that we take the `.prettierignore` file out. Is that right? Totally on board with that.

LINDGRENBA:
> maybe delete prettierignore, or try changing the name to `tokens.js`.

I'll give the latter a try and if it doesn't work, I'll get rid of the `.prettierignore` file.

.prettierignore
laurenmbeatty: yeah...let's just remove it. 👍

PR #50, opened 2021-02-23
add checkbox icons from material ui

### LINKED ISSUE #43: 18. As a user, I would like to have a clear indicator of items purchased
Use round checkboxes similar to iOS devices 

Styling radio buttons as [checkboxes](https://material-ui.com/components/checkboxes/) using [Material UI](https://material-ui.com/components/checkboxes/)

AC:

* Checkboxes should be empty when the item has not been purchased
* Checkboxes should be clearly marked when the item has been purchased

DeeDowns:
## Description

We decided circular checkboxes would give the app a more fun, playful feel. 
And to present a cleaner design, we decided to have a smaller portion of the element change color according to purchase time frame, instead of the entire background. 

This code implements Material UI's Checkbox component and icons to change the square checkboxes to circular ones. 
Both color of the checkboxes and the left border of the list item element holding the shopping list item will change according to how soon the item is expected to be bought again: Soon (red), Kind of soon (orange), Not soon (yellow), Inactive (light grey)    

## Related Issue
Closes #43

Refactors #13 
- Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive

## Acceptance Criteria

- [x] Checkboxes should be empty when the item has not been purchased

- [x]  Checkboxes should be clearly marked when the item has been purchased

## Type of Changes

|    | :bug: Bug fix              |
|    | :100: Add tests            |
|    | :link: Update dependencies |
|    | :scroll: Docs              |

### Before
Screen Shot 2021-02-23 at 6 10 37 PM

### After
Screen Shot 2021-02-23 at 7 36 53 PM

From your terminal, pull down this branch with `git pull origin DD-style-checkboxes` and check that branch out with `git checkout DD-style-checkboxes`
Then type `npm start` in your terminal to launch the app.

OR

Click on the details button below to open the Netlify preview
Screen Shot 2021-02-23 at 9 08 15 PM

Good tokens to test: 
`scope hirsch peat`
`pbs burch sarah`
`mean evict pappy`

### DIFF
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -5,6 +5,9 @@ import Error from './Error';
 import calculateEstimate from './../lib/estimates';
 import { differenceInDays, addDays } from 'date-fns';
 import './List.css';
+import Checkbox from '@material-ui/core/Checkbox';
+import CircleUnchecked from '@material-ui/icons/RadioButtonUnchecked';
+import CircleCheckedFilled from '@material-ui/icons/CheckCircle';
 
 const List = ({ token }) => {
   const { docs, errorMessage, deleteDoc } = useFirestore(token);
@@ -169,8 +172,10 @@ const List = ({ token }) => {
                     item.itemName
                   } ready to purchase ${backgroundColor(item)}`}
                 >
-                  <input
+                  <Checkbox
                     type="checkbox"
+                    icon={<CircleUnchecked />}
+                    checkedIcon={<CircleCheckedFilled />}
                     aria-label="purchased-checkbox"
                     name={item.itemName}
                     id={item.id}
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -128,6 +128,13 @@ const List = ({ token }) => {
     if (daysUntilNextPurchase >= 30) return 'not-too-soon';
   };
 
+  const checkboxColor = {
+    inactive: 'lightgrey',
+    soon: 'red',
+    'kind-of-soon': 'orange',
+    'not-too-soon': 'yellow',
+  };
+
   return (
     <div>
       <h1>List</h1>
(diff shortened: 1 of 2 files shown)


src/components/List.js
```diff
@@ -125,6 +128,13 @@ const List = ({ token }) => {
     if (daysUntilNextPurchase >= 30) return 'not-too-soon';
   };
 
+  const checkboxColor = {
+    inactive: 'lightgrey',
```
DeeDowns, line 132: I initially had the object with all keys in single quotes, not sure when this happened, but now I'm thinking do quotes matter for object keys? 🤔

LINDGRENBA: Hmm... it looks like it's working either way. I'm not sure which way is best-practice. May want to just pick one style and make them all the same? I feel like I've usually seen object keys without quotes, but prob good to get more opinions on this.

LINDGRENBA (approved):
This looks great Dee! The styles here are very nice. Everything works great in the browser. 👍

src/components/List.js
```diff
@@ -125,6 +128,13 @@ const List = ({ token }) => {
...
+    soon: 'red',
+    'kind-of-soon': 'orange',
```
laurenmbeatty, line 134: I think you need the quotes on the keys that have hyphens like this, but I could be wrong. You won't need them on the others, for sure.

DeeDowns: Oh ok, so do you think prettier changed it?  Because I initially had quotes on them all, then noticed it was like this.

veeteeran (approved):
Nice work @DeeDowns. The buttons look great and I really like the way the left border turned out! It looks like there's a merge conflict to resolve. Other than that you're good to merge.

src/components/List.js
laurenmbeatty: Yeah...this is correct. 👍

PR #54, opened 2021-02-24
Rr clear local storage

### LINKED ISSUE #51: 23. As a user, I want to be able to go from an existing list to creating a new list or joining a different list
If the user has a token in local storage, they are not able to navigate back to the home/welcome screen.

AC:
* Create a home button that, when clicked, clears local storage
* User is then taken back to the home/welcome page where they can create a new list or join an existing list

rachelrohrbach:
## Description
This PR adds a navigation link to Home on the List and Add Item pages that gives the user the option to return to the Welcome page after confirming by remove the existing `userToken` from `localStorage` so that the user can create a new list or join a different existing list .  Previously the only way to return to the Welcome page was to manually clear the `userToken` in `localStorage`.  

## Related Issue
closes #51 

## Acceptance Criteria

- [ ] - Create a home button that, when clicked, clears local storage
- [ ] - User is then taken back to the home/welcome page where they can create a new list or join an existing list

## Type of Changes

### Before

beforeHomeLinkScreenshot

### After
removeUserTokenFromStorage

netlifyPreview
2. Otherwise from your terminal, pull down this branch with git pull origin RR-clear-local-storage and check that branch out with git checkout origin RR-clear-local-storage
   * Click the Home link at the bottom of the page while on either the Add Item or List page and confirm that you want to create a new list or join a different existing list. Your user token has been cleared from localStorage and you should now be redirected to the Welcome page to start over with creating a new list or enter a different user token.

### DIFF
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -4,7 +4,7 @@ import { NavLink } from 'react-router-dom';
 const Navigation = ({ confirmDeleteUserToken }) => {
   return (
     <nav>
-      <NavLink to="/" activeClassName="active-link">
+      <NavLink exact to="/" activeClassName="active-link">
         <div className="home-link-text" onClick={confirmDeleteUserToken}>
           Home
         </div>


rachelrohrbach:
So one thing that I can't seem to figure out is why both the Home link is displaying as active when on the list or add item pages.

rachelrohrbach:
@veeteeran @DeeDowns @LINDGRENBA I know we also want to have this functionality on the wireframes with the app logo button and we should easily be able to add that functionality, but I figured I would wait until the icons have been added.

src/components/Navigation.js
```diff
@@ -1,9 +1,14 @@
 import React from 'react';
 import { NavLink } from 'react-router-dom';
 
-const Navigation = () => {
+const Navigation = ({ confirmDeleteUserToken }) => {
   return (
     <nav>
+      <NavLink to="/" activeClassName="active-link">
```
laurenmbeatty, line 7: Try adding the `exact` prop to the NavLink...it seems like then it will only be active if it matches the path exactly. https://reactrouter.com/web/api/NavLink/exact-bool

laurenmbeatty (approved):
Yes...great. Glad that worked.

src/components/Navigation.js
rachelrohrbach: Thanks for that Lauren. The active class works as intended now 😄

DeeDowns (approved):
I like that you gave the user the option to join a new list or not, instead of just clearing local storage.

PR #55, opened 2021-02-24
  Add flex container

### LINKED ISSUE #53: 24. As a user, I want the List view to have the check circles and item name on the right side of the screen and the delete button on the left side of the screen
AC:
* Align check circles and item name to the left side of the screen
* Align delete button to the right side of the screen
* Add lines (a border) between each item

veeteeran:
_For an example of how to fill this template out, [see this Pull Request](https://github.com/the-collab-lab/tcl-3-smart-shopping-list/pull/44)._ 

## Description
This PR includes styling for the list view. The check circle and item name is justified to the left and delete button is justified right

## Related Issue
closes #53 

## Acceptance Criteria

- [x] Align check circles and item name to the left side of the screen
- [x] Align delete button to the right side of the screen
- [x] Add lines (a border) between each item

## Type of Changes

### Before
Screen Shot 2021-02-24 at 5 54 02 PM

### After
Screen Shot 2021-02-24 at 9 03 12 PM

* Otherwise from your terminal, pull down this branch with `git pull origin VT-justify-list-content` and check that branch out with `git checkout VT-justify-list-content`
* Create a new list or join an existing list
* The list shown uses the token `stroll ames knack`

### DIFF
--- a/src/components/List.css
+++ b/src/components/List.css
@@ -17,7 +17,8 @@
 .flex-container {
   display: flex;
   justify-content: space-between;
-  margin-top: 1rem;
+  padding: 1rem 0;
+  width: 385px;
 }
 
 li {
--- a/src/components/List.css
+++ b/src/components/List.css
@@ -13,3 +13,13 @@
 .inactive {
   border-left: 5px solid lightgrey;
 }
+
+.flex-container {
+  display: flex;
+  justify-content: space-between;
+  margin-top: 1rem;
+}
+
+li {
+  border-bottom: 1px solid lightgrey;
+}
--- a/src/components/List.css
+++ b/src/components/List.css
@@ -14,7 +14,7 @@
   border-left: 5px solid lightgrey;
 }
 
-/* .flex-container {
+.flex-container {
   display: flex;
   justify-content: space-between;
   margin-top: 1rem;
@@ -22,4 +22,4 @@
 
 li {
   border-bottom: 1px solid lightgrey;
-} */
+}
--- a/src/components/List.css
+++ b/src/components/List.css
@@ -14,7 +14,7 @@
   border-left: 5px solid lightgrey;
 }
 
-.flex-container {
+/* .flex-container {
   display: flex;
   justify-content: space-between;
   margin-top: 1rem;
@@ -22,4 +22,4 @@
 
 li {
   border-bottom: 1px solid lightgrey;
-}
+} */
--- a/src/components/List.js
+++ b/src/components/List.js
@@ -214,15 +214,6 @@ const List = ({ token }) => {
                           Delete
                         </button>
                       </div>
-                      {/* {item.numberOfPurchases > 0 ? (
-                        <p>
-                          Time until next purchase:{' '}
-                          {getDaysUntilNextPurchase(item)} days. Purchased{' '}
-                          {item.numberOfPurchases} times.
-                        </p>
-                      ) : (
-                        <p>You haven't purchased {item.itemName} yet.</p>
-                      )} */}
                     </li>
                   );
                 })}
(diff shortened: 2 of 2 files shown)


src/components/List.css
```diff
@@ -13,3 +13,13 @@
 .inactive {
   border-left: 5px solid lightgrey;
 }
+
+/* .flex-container {
```
laurenmbeatty, line 17: I think you need to un-comment these. : )

laurenmbeatty:
Material-ui also has some neat Layout components you can use like https://material-ui.com/api/list-item/

src/components/List.css
veeteeran: 😆  You are absolutely correct. I think I need to step away from the screen for a bit

laurenmbeatty (approved):
LGTM 👍

veeteeran:
@laurenmbeatty I initially wanted to stay in the Material UI realm but was having problems getting the spacing. I also tried to use the `box component` which has a flex option. It was probably a case where I defaulted to what I was comfortable using. I definitely want to make use of theming since what I read made it sound like that's a way to apply some overall styling to the app.

veeteeran:
I'll definitely take another crack at it if we for time to refactor or after we demo on Sunday 😄

src/components/List.js
```diff
@@ -138,7 +138,9 @@ const List = ({ token }) => {
   return (
     <div>
       <h1>List</h1>
-
+      {/* <Typography variant="h1" component="h2">
```
chrisdopuch, line 141: Should this commented code be deleted before merging?

veeteeran: Yes, I missed that one. Thanks!

src/components/List.js
```diff
@@ -183,42 +183,46 @@ const List = ({ token }) => {
...
-                      </button>
-                      {item.numberOfPurchases > 0 ? (
+                          {item.itemName}{' '}
+                        </div>
+                        <button onClick={() => confirmDelete(item)}>
+                          Delete
+                        </button>
+                      </div>
+                      {/* {item.numberOfPurchases > 0 ? (
                         <p>
```
veeteeran, line 218: We may use these comments in the future so commenting out for now.

veeteeran, line 218: We may use these comments in the future so commenting out for now.

chrisdopuch: Not a bad idea to keep stuff like that around 👍 
I'd recommend adding a comment explaining that you might use it later. 
Otherwise, someone else might delete it on accident, thinking it was leftover by mistake 🙈

MikeBLambert (approved):
Looks great! 
With regards to the commented out code, remember that with git you can always find old code. For example, if you want to add something back in that was taken out recently, you could always just peruse previously merged PR's to see code that was removed.  :-)

src/components/List.js
```diff
@@ -183,42 +183,37 @@ const List = ({ token }) => {
...
                     >
-                      <Checkbox
-                        type="checkbox"
-                        icon={
-                          <CircleUnchecked
-                            style={{
-                              color: checkboxColor[backgroundColor(item)],
-                            }}
+                      <div className="flex-container">
+                        <div>
```
chrisdopuch, line 187: This approach of using an extra div and the justify-content works OK 👍 
I want to point out that you could also avoid using this extra div, and avoid the stretching of the delete button we're seeing, by using some other flex properties instead.
You could delete this div, and then replace your css with this:
```
.flex-container {
  display: flex;
  align-items: center; // you need this to vertically center the text
  margin-top: 1rem;
}

.flex-container button {
  margin-left: auto; // this "floats" the delete button to the right hand side
}
```

delet button

veeteeran: Cool, thanks for the tip on that. We may need that later on for other buttons! We'll be replacing the delete button on the list view with a trash can icon soon 🗑️

Issue #62, opened 2021-02-25
As a user I want the see header and navigation elements clearly delineated from the main part of the app


DeeDowns:
We came up with styling to clearly outline the main part of the app from the logo/title and bottom navigation. 
Styling should reflect that of our wireframe:
<img width="708" alt="Screen Shot 2021-02-25 at 5 07 10 PM" src="https://user-images.githubusercontent.com/47228065/109226470-cd64e380-778c-11eb-92bc-3579351f71a7.png">

AC :
- There is a header element to separate the title and logo from main part of the app
- There is a border to separate navigation elements from main part of the app 
- The header background color and border color should be the same
- The `List` page and `Add Item` page will have the same styles applied

PR #63, opened 2021-02-25
create Header component

DeeDowns:
## Description

Adds header and top border with styling to outline the main part of the app

## Related Issue

closes[#62]

## Acceptance Criteria

- [x] There is a header element to separate the title and logo from main part of the app

- [x] There is a border to separate navigation elements from main part of the app

- [x] The header background color and border color are be the same

- [x] The `List` and `Add Item` pages have the same styles applied to them

## Type of Changes

|    | :bug: Bug fix              |
|    | :hammer: Refactoring       |
|    | :100: Add tests            |
|    | :link: Update dependencies |
|    | :scroll: Docs              |

### Before

header-nav

### After

header-nav-styled

From your terminal, pull down this branch with `git pull origin DD-header-nav-outline` and check that branch out with `git checkout DD-header-nav-outline`
Then type `npm start` in your terminal to launch the app.

OR
Click on the details button below to open the Netlify preview
Screen Shot 2021-02-23 at 9 08 15 PM

- Any token will do (`calf bebop styx`)

- Navigate to the Add Item or List page to view the header and styling above the navigation elements

### DIFF
--- a/src/components/AddItem.js
+++ b/src/components/AddItem.js
@@ -2,6 +2,8 @@ import React, { useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { db } from '../lib/firebase';
 import Error from './Error';
+import Header from './Header';
+import '../styles/AddItem.css';
 
 const initialFormState = {
   itemName: '',
@@ -54,7 +56,8 @@ const AddItem = ({ token }) => {
   };
 
   return (
-    <div>
+    <div className="add-item-container">
+      <Header />
       <h1>Add Item</h1>
       <form onSubmit={handleSubmit}>
         <label>
@@ -105,7 +108,7 @@ const AddItem = ({ token }) => {
         </fieldset>
         <button type="submit">
           <span>
-            Add Item<i class="fas fa-cart-plus"></i>
+            Add Item<i className="fas fa-cart-plus"></i>
           </span>
         </button>
       </form>
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import '../styles/Header.css';
 
 const Header = () => {
   return (
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -11,8 +11,3 @@ const Header = () => {
 };
 
 export default Header;
-
-// There is a header element to separate the title and logo from main part of the app
-// There is a border to separate navigation elements from main part of the app
-// The header background color and border color should be the same
-// The List page and Add Item page will have the same styles applied
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -1,0 +1,17 @@
+import React from 'react';
+
+const Header = () => {
+  return (
+    <header>
+      <p>logo</p>
+      <h1>Smart Shopper</h1>
+    </header>
+  );
+};
+
+export default Header;
+
+// There is a header element to separate the title and logo from main part of the app
+// There is a border to separate navigation elements from main part of the app
+// The header background color and border color should be the same
+// The List page and Add Item page will have the same styles applied
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -1,10 +1,15 @@
 import React from 'react';
 import '../styles/Header.css';
+import homeIcon from './../img/groceries.svg';
 
 const Header = () => {
   return (
     <header>
-      <p>logo</p>
+      <img
+        src={homeIcon}
+        className="homeIcon"
+        alt="colorful grocery bag logo and home link"
+      />
       <h1>Smart Shopper</h1>
     </header>
   );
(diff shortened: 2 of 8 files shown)


src/components/List.css
```diff
@@ -1,3 +1,11 @@
+.list-container {
```
DeeDowns, line 1: I had to widen the top level div for the List and Add Item page so the header would span the entire width of the page

src/components/List.js
```diff
@@ -136,8 +137,8 @@ const List = ({ token }) => {
   };
 
   return (
-    <div>
-      <h1>List</h1>
+    <div className="list-container">
+      <Header />
```
DeeDowns, line 141: I added the Header component to the List and Add Item components and not the App component, because the Welcome page will have its own styling and won't include the header or navigation elements.

src/styles/Header.css
```diff
@@ -0,0 +1,16 @@
...
+  justify-content: center;
+  align-items: center;
+}
+
+h1 {
+  width: 80%;
+  font-size: 2.4rem;
+}
+
+header .homeIcon {
```
DeeDowns, line 14: Let me know if you think the size of the logo and/or title font size should change

src/components/AddItem.js
```diff
@@ -105,7 +108,7 @@ const AddItem = ({ token }) => {
         </fieldset>
         <button type="submit">
           <span>
-            Add Item<i class="fas fa-cart-plus"></i>
+            Add Item<i className="fas fa-cart-plus"></i>
```
laurenmbeatty, line 111: Thank you!

src/components/Header.js
```diff
@@ -0,0 +1,18 @@
...
+import '../styles/Header.css';
+import homeIcon from './../img/groceries.svg';
+
+const Header = () => {
+  return (
+    <header>
+      <img
+        src={homeIcon}
+        className="homeIcon"
+        alt="colorful grocery bag logo and home link"
```
laurenmbeatty, line 11: excellent

laurenmbeatty:
@DeeDowns Use the dev tools to give a mobile view. This is what I see on the List page:

image

Just need to toggle here:
image

laurenmbeatty:
Having said that, if I navigate to the Add Item page, and then back to the list, then it looks OK. You all might want to consider putting the list in a scrolling container, so that the navigation is always at the bottom of the screen.

Oh, I see why. Something in the list items has a width of 385px...could just make the width of the ul be like 90% or something, and then get rid of that specific width.

laurenmbeatty:
Then it would look like this:
image

laurenmbeatty (approved):
Looks cute!  Probably want to create a separate issue for the list items.

DeeDowns:
> Having said that, if I navigate to the Add Item page, and then back to the list, then it looks OK. You all might want to consider putting the list in a scrolling container, so that the navigation is always at the bottom of the screen.
> 
> Oh, I see why. Something in the list items has a width of 385px...could just make the width of the ul be like 90% or something, and then get rid of that specific width. Would also need to get rid of the padding on the ul...unordered lists have 15px of left padding built in.
> 
> You could just create a separate issue for that, since it's not part of this ticket. 👍

Oh ok so I changed the `.flex-direction` width to 100% and added the `ul` width to be 25% and that fixed the white showing when you resize to mobile view. I'm not going to push it, but will make a note for when we get to media queries and responsive styling.
I think this was from another issue, I normally use % instead of px for width. Is it better to use one over the other?

MikeBLambert (approved):
Header looks great! 
You may want to consider rendering the Header component making the \<Header /\> component part of the \<App /\> and only rendering it if there is a token. 
That, or you could make something like a \<Page /\> component, which renders a div that has the css that's duplicated for both the .list-container and the .add-item-container, with the \<Header /\> 
It would look something like 
const Page = ({children}) => 
\<div className="page"\>
\<Header /\>
    {children}
    \</div\>

Then, rather than a \<div /\> container in \<List /\> and \<AddItem /\> you use your \<Page /\> component and...voila! 
Just an idea. 
Again, what you've got here looks great!

laurenmbeatty:
> Oh ok so I changed the `.flex-direction` width to 100% and added the `ul` width to be 25% and that fixed the white showing when you resize to mobile view. I'm not going to push it, but will make a note for when we get to media queries and responsive styling.
> I think this was from another issue, I normally use % instead of px for width. Is it better to use one over the other?

Yeah, flexible whenever possible, for sure. Sometimes we need fixed size, but i usually try to go for flexible.

src/components/List.css
LINDGRENBA: Another way to possibly tackle that would be to create a Layout that wraps around App and put the header and footer in the layout. But I think if the header issue in the mobile view is resolved based on what you and Lauren talked about below then that's unnecessary.

LINDGRENBA (approved):
It looks really good Dee! Everything is consistent and I like that you figured out how we can resolve the header issue in mobile view, but made the decision to leave that for the issue related to the mobile view!

PR #67, opened 2021-02-26
  Add textField, spacing between rest button

### LINKED ISSUE #56: 25. Clean up spacing of search filter in List view, combine fix for accessibility
There are two main accessibility items on the List page:

1. The Delete icon buttons and the "reset" button are "empty."
2. The Checkboxes don't have associated form labels.
image

Definition of done:
* These issues have been remediated, and a Lighthouse score of 100 is achieved, or "Wave" gives no errors.

This is a great article on how to handle icon/empty buttons: https://www.sarasoueidan.com/blog/accessible-icon-buttons/

AC:

* create space between label, input element, and button

veeteeran:
## Description
This PR adds styles using the Material UI `textField` component for the search box.

## Related Issue

closes #56

## Acceptance Criteria

- [x] Create space between label, input element, and button
- [x] Fix accessibility for `reset` button and `delete` icon

## Type of Changes

|  ✓ |:hammer: Refactoring       |

### Before
Screen Shot 2021-02-26 at 1 18 13 PM

### After
Screen Shot 2021-02-26 at 1 10 47 PM

* Otherwise from your terminal, pull down this branch with `git pull origin VT-list-search-filter-spacing` and check that branch out with `git checkout VT-list-search-filter-spacing`
* Create a new list or join an existing list
* The list shown uses the token `stroll ames knack`


rachelrohrbach (approved):
Looks good! I think this brings more uniformity to the app and I like that the input field has the nice rounded corners keeping with our "theme".

laurenmbeatty:
I would think about adding more space between the Add item button and the search bar....could be a separate issue. Maybe wait and see until the header is in place, too to evaluate it altogether.

image

laurenmbeatty (approved):
LGTM 👍

src/components/List.js
```diff
@@ -4,8 +4,8 @@ import { db } from '../lib/firebase';
 import Error from './Error';
 import calculateEstimate from './../lib/estimates';
 import { differenceInDays, addDays } from 'date-fns';
-import './List.css';
-import Checkbox from '@material-ui/core/Checkbox';
+import '../styles/List.css';
```
chrisdopuch, line 7: @veeteeran it looks like you moved the List.css file down into a styles directory.
However, I'm not seeing the file deleted/renamed in this PR. Is the old List.css still hanging around?

Also, I'd be curious to know what your thought process was around moving the file. Co-locating your css files with the component that they style is actually a pretty common practice in modern development, as modularization allowed us to split our CSS files up into component-based chunks, instead of one big global CSS file for the whole app. 
Not saying you shouldn't do this necessarily - I'm just curious what your reasoning was.

src/components/List.js
```diff
@@ -157,17 +157,17 @@ const List = ({ token }) => {
...
-              <input
-                type="text"
-                name="search-bar"
-                id="search-bar"
-                placeholder="Start typing here..."
-                value={searchInput}
-                onChange={handleSearchChange}
-              />
-              <input type="reset" onClick={handleClear} />
+              <Box display="flex" alignItems="center">
```
chrisdopuch, line 160: Nice use of the Material-UI components! 👍 
Now that the input is taller, it really calls attention to the fact that the reset button is much shorter than the text input.
Have you thought about swapping out the reset button for a Material-UI Button component, too?

src/components/List.js
```diff
@@ -157,17 +157,17 @@ const List = ({ token }) => {
...
+                <TextField
+                  name="search-bar"
+                  id="search-bar"
```
chrisdopuch, line 163: Is this id being used? I don't see it being referenced in the List.css file.

chrisdopuch (approved):
Nice improvement! I had some questions around the CSS moving, and a possibly unused ID. I was also curious if you were going to try to port the button to Material-UI, too, but that can be a stretch goal 👍

src/components/List.js
```diff
@@ -157,17 +157,17 @@ const List = ({ token }) => {
...
+                  label="Start typing here..."
```
DeeDowns, line 164: Would it be better to label the text input so the user knows its a search bar, like "filter items" or "search". Unless we are adding the magnifying icon to the text input, right now it is a bit ambiguous as to what the text input is for. I mean we know because we created it, but would a user know that this is to filter list items?

src/components/List.js
veeteeran: Hi @chrisdopuch we to move our css files into a styles folder since we thought the organization was better? 😄  Just personal preference on our part. It's good to know that it's common practice to keep the css and components together

src/components/List.js
veeteeran: @chrisdopuch You read out minds!

src/components/List.js
veeteeran: 👍🏽  We can delete the id there

src/components/List.js
veeteeran: @DeeDowns great suggestion. I was thinking it should be named something different. I think changing the placeholder to `filter list` makes more sense too

Issue #40, opened 2021-02-20
As a user I want the application to have a nice color palette so it is fun to look at


LINDGRENBA:
Users want applications that are pleasing to look at and if the application is food related, colors should not be unappetizing. 

AC : 
* Colors and contrast must meet accessibility guidelines / pass accessibility tests
[WCAG Color Contrast Testing Site](https://accessible-colors.com/)
* Colors and whitespace are used to create a clean, fun theme

LINDGRENBA:
I updated this to just address the implementation of the material UI theme for now, but will check in with the team to finalize where the classes / colors should be specifically implemented.

PR #71, opened 2021-02-27
Bl add color theme

LINDGRENBA:
## Description
This PR will add a Material UI Theme, implement the useStyles and makeStyles functions and make it possible to set colors within components using simple references such as `red`, `blue` and `salmon`, etc. eliminating the need to use hex codes and making it easy to implement consistent styling accross the entire application.

## Related Issue
[#40 ](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/40)

## Acceptance Criteria
- [ ] Colors and contrast must meet accessibility guidelines / pass accessibility tests
[WCAG Color Contrast Testing Site](https://accessible-colors.com/)
- [ ] Material UI Theme is implemented so that colors are easily accessed and remain consistent throughout the application.

## Type of Changes

### Before
before-theme

### After
after-theme

## How to use the theme colors in your components
Access the theme colors by setting the `className` attribute on an element and setting its value to `{classes.colorYouWantToAccess}`, see example below : 
using-mui-theme

*The Material UI Theme is not currently applied to any specific element, so if you would like to test this feature:*
* From your terminal, pull down this branch with `git pull origin BL-add-color-theme` and check that branch out with `git checkout BL-add-color-theme`
* `npm start` to launch the app
* Start a list and navigate to the List view
* Copy the following: `<h1 className={classes.salmon}>IS THIS WORKING</h1>`
* Paste the line of code above inbetween the `<Header />` and the `{loading ? ` , depicted in the image below as line 163
* list-view

* Refresh the list page to confirm that `THIS IS WORKING` appears and the color is salmon '#f98365'


src/components/Welcome.js
```diff
@@ -1,8 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import TokenForm from './TokenForm';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles((theme) => ({
```
laurenmbeatty, line 6: I love this function. I think it would be nice to export this `useStyles` function from the theme, then just import it and use in the component, rather than repeating it in every file.

So in the component files, you would just need the import statement, and the `const classes = useStyles()`

src/styles/Theme.js
```diff
@@ -0,0 +1,29 @@
...
+    salmon: {
+      main: '#f98365',
+    },
+    yellow: {
+      main: '#F9de59',
+    },
+  },
+});
+
+console.log('theme file ', theme);
```
laurenmbeatty, line 23: maybe can get rid of this and the commented out code below. 👍

MikeBLambert (approved):
So excited for how this is coming together!

src/components/Welcome.js
LINDGRENBA: Yes! Thank you. I was trying to wrap my head around how to refactor and just couldn't get there. Thank you for the suggestion! It's way more DRY this way.

src/styles/Theme.js
LINDGRENBA: There is one issue with color contrast - if you enter an invalid token on the welcome screen, the WAVE plugin registers poor contrast between the form's yellow background and the red error message. Due to my internet issues, @DeeDowns has kindly offered to resolve this error.

Issue #83, opened 2021-03-02
As a user, I want to be able to invite others to join my list


LINDGRENBA:
Giving the user access to their list token will allow them to share their token and invite others to join their list.

AC : 

- [ ] The user's current token is displayed above their list
- [ ] The user sees a message instructing them on the purpose of the three word token 
- [ ] The user can select an icon that enables `copy-to-clipboard` for their token

PR #84, opened 2021-03-14
Bl change firestore

LINDGRENBA:
## Description

This PR will do the following three things:
1. Update the firestore config keys
2. Show the user their 3 word token, so they can share it and invite other users to join their list
3. Add meta data to allow open graph protocol functionality - eg: sharing the application url will also share an image preview of the application

## Related Issue
closes [#83](https://github.com/the-collab-lab/tcl-17-smart-shopping-list/issues/83)

## Acceptance Criteria
- [ ] The user's current token is displayed above their list
- [ ] The user sees a message instructing them on the purpose of the three word token 
- [ ] The user can select an icon that enables `copy-to-clipboard` for their token

## Type of Changes

### Before
before-adding-token

### After
after-adding-token

* Otherwise from your terminal, pull down this branch with `git pull origin BL-change-firestore` and check that branch out with `git checkout BL-change-firestore`
* Run `npm start` in your terminal to launch the application in the browser 
1. Create a new list
2. Confirm that once you create a new list, the load screen is replaced by the list view - *(this will confirm that the new firestore keys are working as expected)*
3. Confirm that you see the message `Invite a friend to join your list by sharing this three word token: token token token` at the top of your list view - *(the three word token displayed at the end of the message will be different for each list)*
4. Click on the clipboard icon, open a text document of your choice (google docs, microsoft word, etc.) and paste *(right click and paste, ctrl+v or cmd+v)* to confirm that your three word token was successfully copied to your clipboard


public/index.html
```diff
@@ -2,13 +2,21 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
```
LINDGRENBA, line 4: I'm not actually sure how to confirm that this is working until we're able to set up and deploy to a new Netlify account... does anyone know if there's a way to test it at this time?

src/lib/firebase.js
```diff
@@ -5,12 +5,12 @@ import 'firebase/firestore';
```
LINDGRENBA, line 5: I think that at some point it would be good to take @MikeBLambert 's advice and store credentials in a `.env` file, but I just wanted to get this set up before we lost our read/write permissions - I def want to avoid getting a large bill from firestore 😄

PR #85, opened 2021-03-20
Test automated deploys to Firebase

segdeha:
If this works, we should get a preview deploy link in the PR. When the PR is merged, the site should go live on Firebase hosting.


segdeha:
Merging this should make the app live on Firebase hosting. 🤞

PR #86, opened 2021-04-06
Bl update hosting

LINDGRENBA:
## Description
This PR will set up CI/CD for our project repo, giving the team access to the build pipeline which will allow them to more easily make changes and updates. 

## Related Issue

no related issue

## Acceptance Criteria

- [ ] Preview deploys should work when new PRs are opened.
- [ ] Merging PRs to main should deploy to production.

## Type of Changes

|  ✓ | 🏠 : Set up CI/CD and hosting for project repo         |

### Before

not applicable

### After

not applicable

Check below to confirm that the preview deploys are working. It should appear as :

image

Visit the URL and confirm that the site works as expected.

Once the PR is merged, confirm that the production deploys are working as expected.

Once merged, as stated in the document [How to set up CI/CD for project repos](https://www.notion.so/How-to-set-up-CI-CD-for-project-repos-3dc27b787d344218b74060d47532b0b6) 
```Note: You can check the Actions tab in the repo to see the progress/status of the workflows.```

### DIFF
--- a/.github/workflows/firebase-hosting-merge.yml
+++ b/.github/workflows/firebase-hosting-merge.yml
@@ -11,12 +11,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: npm install && npm run build
+      - run: npm ci && npm run build
       - uses: FirebaseExtended/action-hosting-deploy@v0
         with:
           repoToken: '${{ secrets.GITHUB_TOKEN }}'
-          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_TCL_17_SHOPPING_LIST }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_SMART_SHOPPER_E95D9 }}'
           channelId: live
-          projectId: tcl-17-shopping-list
+          projectId: smart-shopper-e95d9
         env:
           FIREBASE_CLI_PREVIEWS: hostingchannels
--- a/firebase-debug.1.log
+++ b/firebase-debug.1.log
@@ -1,0 +1,47 @@
+[debug] [2021-04-06T03:53:14.012Z] ----------------------------------------------------------------------
+[debug] [2021-04-06T03:53:14.015Z] Command:       C:\Program Files\nodejs\node.exe C:\Users\blindgren\AppData\Roaming\npm\node_modules\firebase-tools\lib\bin\firebase.js init
+[debug] [2021-04-06T03:53:14.015Z] CLI Version:   9.8.0
+[debug] [2021-04-06T03:53:14.015Z] Platform:      win32
+[debug] [2021-04-06T03:53:14.015Z] Node Version:  v10.15.1
+[debug] [2021-04-06T03:53:14.017Z] Time:          Mon Apr 05 2021 20:53:14 GMT-0700 (Pacific Daylight Time)
+[debug] [2021-04-06T03:53:14.017Z] ----------------------------------------------------------------------
+[debug] 
+[debug] [2021-04-06T03:53:14.029Z] > command requires scopes: ["email","openid","https://www.googleapis.com/auth/cloudplatformprojects.readonly","https://www.googleapis.com/auth/firebase","https://www.googleapis.com/auth/cloud-platform"]
+[debug] [2021-04-06T03:53:14.029Z] > authorizing via signed-in user
+[info] 
+     ######## #### ########  ######## ########     ###     ######  ########
+     ##        ##  ##     ## ##       ##     ##  ##   ##  ##       ##
+     ######    ##  ########  ######   ########  #########  ######  ######
+     ##        ##  ##    ##  ##       ##     ## ##     ##       ## ##
+     ##       #### ##     ## ######## ########  ##     ##  ######  ########
+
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: .github/workflows/firebase-hosting-merge.yml [hunk 1]
```diff
@@ -11,12 +11,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: npm install && npm run build
+      - run: npm ci && npm run build
```
laurenmbeatty, line 14: Did you keep the default option on this one? I think this might have to stay as `npm install && npm run build`. There isn't an `npm ci` script in our package.json that I can see.

laurenmbeatty: I mean, on the other hand it seems like the hosting is working 😂

.github/workflows/firebase-hosting-merge.yml
```diff
@@ -11,12 +11,12 @@ jobs:
...
       - uses: FirebaseExtended/action-hosting-deploy@v0
         with:
           repoToken: '${{ secrets.GITHUB_TOKEN }}'
-          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_TCL_17_SHOPPING_LIST }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_SMART_SHOPPER_E95D9 }}'
           channelId: live
-          projectId: tcl-17-shopping-list
+          projectId: smart-shopper-e95d9
```
laurenmbeatty, line 20: 👍 

I just set this up on a project I'm doing...it's pretty slick.

functions/index.js
```diff
@@ -0,0 +1,8 @@
+const functions = require('firebase-functions');
+
+// // Create and Deploy Your First Cloud Functions
```
laurenmbeatty, line 3: Do you still need this? Is this auto-generated?

firebase-debug.1.log
```diff
@@ -0,0 +1,47 @@
+[debug] [2021-04-06T03:53:14.012Z] ----------------------------------------------------------------------
```
laurenmbeatty, line 1: I think this could probably be deleted? I don't think it needs to be in the project.

laurenmbeatty (approved):
Just a couple of questions, but it seems like it's working!

.github/workflows/firebase-hosting-merge.yml
LINDGRENBA: Oh my goodness, for some reason I didn't see any notifications that anyone had reviewed 😄 so I'm just seeing your comments here. Apologies!!!

LINDGRENBA: It's good that it seems like it's working! When following along with the [Notion doc](https://www.notion.so/How-to-set-up-CI-CD-for-project-repos-3dc27b787d344218b74060d47532b0b6) @segdeha shared, in step 2, substep 10 for initializing the project, I did specify that `npm install && npm run build` should run before every deploy, so maybe `npm ci` is something build into the process? I did a brief search for a reference to it in the firebase docs, but with no luck.

functions/index.js
LINDGRENBA: Hmm... not sure about that - if we're still using it. I'll look into it.

firebase-debug.1.log
LINDGRENBA: or maybe added to the .gitignore? which do you think would be better?

laurenmbeatty: I'm not 100% sure, but I'm 99.9% sure you can just delete it. It looks like a file that got automatically. created when something went wrong. Safe to delete! 👍

.github/workflows/firebase-hosting-merge.yml
laurenmbeatty: I think it will be fine this way, I just looked up the difference between npm ci and npm install [here](https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci#:~:text=npm%20ci%20(named%20after%20Continuous,it%20will%20throw%20an%20error.), so I think it should be fine. Worst case scenario, we just do the firebase hosting set-up process again.

LINDGRENBA: I think you're right @laurenmbeatty - thank you for linking to that SO. 👍

firebase-debug.1.log
LINDGRENBA: That's so weird... I totally agree with you - don't think we need it, but on opening my feature branch of the project, I can't find this log anywhere 🤔

functions/index.js
LINDGRENBA: It looks like this [Get Started](https://firebase.google.com/docs/functions/get-started) firebase doc mentions this. It looks like we would only need it if we wanted to build or connect cloud services to the project, if I'm understanding that correctly. Does that sound right? If that's the case then we could probably get rid of it, or we could leave it in case we decide to get fancy in the future... 😄 What do you think @laurenmbeatty ?

LINDGRENBA:
Yikes. Sorry it took me 3 months to merge this. I am def. embarrassed. I hope it didn't inconvenience anyone. Thanks again @laurenmbeatty for your review and your comments. 😄

